PR #16, opened 2014-09-26
Add react-dragged class to elements that have been dragged

lmb:
This PR allows elements to stay in normal flow until they are actually dragged.


STRML:
The underlying issue was fixed in 611c895d0c9f91277b4e73c4fc96d05a36ab24ed.

PR #31, opened 2014-12-28
New features: `moveOnStartChange`, event updates, and more

STRML:
This is kind of a large PR so please let me know if you'd like it split up.

This is some of the work I did on Draggable to help it play nice with my new [grid project](https://strml.github.io/react-grid-layout/examples/1.html).

The changes are:
- JSHint fixes
- Add `moveOnStartChange`. This flag tells the Draggable to update its `left` and `top` when the `start` param changes. This is used in React-Grid-Layout to move the Draggables when the layout updates.
- Added the DOM Node to the fired event. I use this in RGL for a few reasons and it's not reliably available via `e.target`, and `e.currentTarget` is set incorrectly for some reason. I thought it would be helpful to have.
- Added a class while dragging the element.
- Added a class to the body while dragging to disable user-select. Without it, it's easy to get blue highlights all over your page.

Let me know what you think - if possible it would be great to merge all of the changes rather than split them up.

Thanks

### DIFF
--- a/lib/draggable.js
+++ b/lib/draggable.js
@@ -500,7 +500,7 @@ module.exports = React.createClass({
 		// This makes it flexible to use whatever element is wanted (div, ul, etc)
 		return React.addons.cloneWithProps(React.Children.only(this.props.children), {
 			style: style,
-			className: 'react-draggable',
+			className: 'react-draggable' + (this.state.dragging ? ' react-draggable-dragging' : ''),
 
 			onMouseDown: this.handleDragStart,
 			onTouchStart: function(ev){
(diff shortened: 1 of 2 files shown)


mzabriskie:
So I just refactored this component on a separate branch. I am trying to decide what is the best way to handle merging all this. Let me pick through your PR and see what makes sense.

STRML:
Sure. I'm running a fork in the react-grid-layout project, would be nice
to get rid of the github dep though; maybe will fork and rename if we
can't get it merged. I think they're useful additions though. Sorry
they're not more split up but I was working fast.

STRML:
All of these updates are finally merged.

PR #36, opened 2015-02-23
Add drag bound property

mgechev:
Added a bound property. It allows to bound the draggable element by it's parent (using value `"parent"`) or given rectangle (by specifying `{ left: LEFT_BOUND, right: RIGHT_BOUND, bottom: BOTTOM_BOUND, top: TOP_BOUND }`).

I tried to limit the changes as much as possible, so followed the tab indentation you have, although I'd suggest to add `.jscsrc` with stricter configuration.

I also wasn't able to add better tests, only default property value check. When I tried to trigger `mouseMove`, using `TestUtils.Simulator` the `mousemove` callback wasn't called. I didn't have enough time to make sure about the exact reason but I think it is because of lack of `reactid` of the `window` object, so the event wasn't propagated internally.


STRML:
This is great - I will manually merge this in prep for `0.5.0`.

PR #39, opened 2015-03-19
Add the option to reset the element

andreimelinte:
Add the option to reset the element's client X and Y position to the start position when drag ends.


STRML:
Could you tell me more about the use case for this? I am writing a 0.5.0 release with some breaking changes and I am considering including this type of functionality.

mweststrate:
I am currently trying to achieve that it is possible to drag a clone of an object (to drag something from a  toolbox into a canvas). That seems to be a perfect case for this scenario; you want to item to be put back into the toolbox after finishing a drag, so that you can drop it another time

mweststrate:
Also, when returning `false` from either the `onDrag` (or `onStop`!) event, I would expect the element to return to its original position in all cases.

STRML:
It may make more sense to simply reset `start` to `[0,0]` and set `moveOnStartChange={true}` when the `onStop` handler fires.

STRML:
Use `<DraggableCore>` from `v1.0.0` if you need this.

PR #42, opened 2015-03-21
Require react not react/addons

raysuelzer:
This fixes the issue of the module not working in webpack (at least for me).


STRML:
This is imo a bad idea because we're using classSet, but classSet is deprecated too so I will adjust and move to [classNames](https://www.npmjs.com/package/classnames).

PR #45, opened 2015-04-07
Browserify transforms should be regular dependencies.

seansfkelley:
This prevents CoffeeScript-based projects that don't already include reactify from throwing exceptions while running Browserify.


STRML:
bc11f76 introduces use of the `"browser"` field. See https://github.com/substack/node-browserify#browser-field

PR #47, opened 2015-04-07
Only consider a drag properly started if it heads along the specified axis.

seansfkelley:
This fixes an issue where a drag along e.g. the x axis would prevent all vertical scroll events. This is bad when your draggable elements are inside a list, such as you might do when emulating iOS's drag-over-to-delete idiom.

Now, a drag is ended immediately if it starts by moving in the wrong direction. Additionally, all move events have their propagation stopped to prevent scrolling.

Two open questions I have here:
1. Should this be configurable? Wouldn't be too hard, but I don't want to clutter the public interface with too many options (part of the reason I chose this of react-dnd is because the latter is way too complicated for my uses).
2. Should I stopPropagation on all the events, including start and end? Seems weird to leak them but not touchmove.


STRML:
If you still want to merge this (and 0.8.0 hasn't fixed this use case), could you submit a jsfiddle that demonstrates the error?

seansfkelley:
So this issue only really comes up on touch screens, where the touching the screen can both drag an element and scroll the page. Unfortunately I wasn't able to get it to behave like that even with touch emulation turned on in my browser, so the repros outlined below are all for touch screens. It turns out that 0.4.0 and 0.8.1 behave very differently here. I've outlined the three behaviors below (including my change). For each one just touch the red box in the preview window and try dragging it around. (It has `{axis: 'x'}` so it's locked to the horizontal). You may want to zoom in a bit too, to give some space to see the scrolling happen.

For reference, I'm running on and iPhone (Mobile Safari, iOS 8.3). I have to start the drag gesture diagonally otherwise iOS will lock my scroll direction along the major axis, which interferes with these repros.
- [In the 0.4.0 version](http://jsfiddle.net/66ho6eb8/5/), the draggable swallows all the events, so even if the draggable is locked on one axis (here, the X-axis), you can't scroll vertically.
- [In the 0.8.1 version](http://jsfiddle.net/66ho6eb8/1/), the draggable passes through all events, so not only do you scroll vertically (desirable), you also scroll horizontally _while simultaneously dragging horizontally_ (not desirable).
- [In my change against 0.4.0](http://jsfiddle.net/66ho6eb8/6/), the draggable surrenders control if you start dragging vertically and stays frozen, but swallows scroll events if you start dragging horizontally so the page stays still. I think this feels more natural.

Also, it feels a bit weird that if I start a scroll gesture, if I'm sloppy about it (wavering my finger perpendicularly to the major axis of motion), the element follows me anyway, which is why I opted for scrolling and dragging to be mutually exclusive.

STRML:
Thanks for the fiddles. I'm going to go through them with a mobile device on `v1.0.0` and see if this is still manifesting. Dragging on a touch device is always a bit fiddly but I'd like this project to do its best to make it work.

seansfkelley:
This can be done using the start/drag/end hooks in 1.1.1.

brettlyne:
@seansfkelley 

> This can be done using the start/drag/end hooks in 1.1.1.

How? Seems like event.preventDefault() is called by default. Not sure how my onDrag handler can un-preventDefault. Thanks.

seansfkelley:
@Morlem it's been a long time since I've used this code, but IIRC, 1.1.1 passes through all events (not sure about newer versions), so all you have to do is `preventDefault`. If react-draggable has since changed to call `preventDefault` by default, then this issue is probably still relevant.

PR #53, opened 2015-05-13
Move brief API documentation out of example

markthethomas:
Really love this component and thought it might be helpful for other people if the props for a `<Draggable/>` were more clearly documented. I don't usually look in code examples for documentation about a given API and it might help people clearly see how to clearly use the module while still including an easy-to-get usage example. Open to thoughts / feedback / &c.!


STRML:
Thanks!

markthethomas:
no prob!

PR #62, opened 2015-06-12
Only react to changes in start param if it has actually changed

benjeffery:
Hi,
Loving react-draggable, thanks!
I came across an issue where a component needed to re-render, sending an unchanged "start" prop to a Draggable while a drag was happening. Adding logic to not reset the state when "start" was unchanged fixed this. I think this behaviour is sensible as a default, hence this pull request.

Thanks again!

### DIFF
--- a/lib/draggable.js
+++ b/lib/draggable.js
@@ -467,8 +467,10 @@ module.exports = React.createClass({
   },
 
   componentWillReceiveProps: function(newProps) {
-    // React to changes in the 'start' param.
-    if (newProps.moveOnStartChange && newProps.start) {
+    // React to changes in the 'start' param, but only if it has changed.
+    if (newProps.moveOnStartChange && newProps.start &&
+      this.props.start && this.props.start.x !== newProps.start.x &&
+      this.props.start.y !== newProps.start.y) {
       this.setState(this.getInitialState(newProps));
     }
   },


STRML:
Hey, thanks for this.

I mean to rework how this module behaves; I've found that while developing
React-Grid-Layout, the internal state that <Draggable> manages is becoming
unwieldy when I want to move items manually. There should be an easy way to
both encapsulate <Draggable> in a wrapper that manages its own draggable
state, and move it via external forces as well.

I will likely move <Draggable> into two separate exported modules, likely
<Draggable> and <DraggableCore>. It will be a clean break from the current
API, unfortunately. If backcompat is possible I will try but I can't
guarantee anything.

Thanks for the PR, (I appreciate it a lot) but it likely won't make it in
because this rewrite is pending.

mweststrate:
I recognize that it is currently hard to externally influence the draggable
state. However, a simple yet effective work-around for us was to reset the
internal state of the draggable by assigning a new 'key' property which
causes react to throw away the old draggable (and thus its state) and
creating a new one. At that point you can also assign a new base position
for example, while avoiding that the last relative drop position to be
added to it.

If that is an acceptable work around, it might be worth mentioning it in
the readme.

STRML:
That is very interesting, thank you for the suggestion. I will try it out
with RGL and if it works, we can postpone / cancel the rewrite and just
document this. Thanks!

benjeffery:
Understand about the rewrite. Wouldn't bother me if you broke the API to make it cleaner rather than compromise it by being backwards compatible.

I don't think that workaround mentioned by @mweststrate will work in my case - I want to keep the state.

STRML:
See `<DraggableCore>` in `v1.0.0`.

Issue #63, opened 2015-06-13
Percentage based drag system


AndrewRayCode:
I'm trying to create resizable split panes. If a draggable component (the handle between panes) is set to a % offset from its parent, it's much more convenient. I'm not sure I can fanangle it with a transform: translate setup

joeporpeglia:
I have a similar use case and I'm using `DraggableCore` to handle it. Just manually set the `left` and `top` styles as a percentage on `DraggableCore`'s child.

phillfer:
@joeporpeglia Could you give an example on how you did that?

damassi:
ping @joeporpeglia

designcrumble:
For anyone still looking for an answer, this might be helpful:
I am storing x, y positions of the item relative to its parent/bounds container. Then just setting position absolute on the Child and passing `top` and `left` styles after converting the `float` to %. 
``` javascript
<Draggable
  key={i}
  axis="both"
  handle=".handle"
  position={null}
  cancel=".content-is-editable"
  // grid={[25, 25]}
  scale={1}
  disabled={props.disabled}
  {...dragHandlers}
>
  <div
    className="handle cursor-move absolute"
    id={item.id}
    datatype={item.item_type}
    style={{ left: `${item.x * 100}%`, top: `${item.y * 100}%` }}
  >
    <div className="item item--media text-serif text-styled-500">
      {item.content}
    </div>
  </div>
</Draggable>;

```

isaac-wahba:
> For anyone still looking for an answer, this might be helpful: I am storing x, y positions of the item relative to its parent/bounds container. Then just setting position absolute on the Child and passing `top` and `left` styles after converting the `float` to %.
> 
> ```js
> <Draggable
>   key={i}
>   axis="both"
>   handle=".handle"
>   position={null}
>   cancel=".content-is-editable"
>   // grid={[25, 25]}
>   scale={1}
>   disabled={props.disabled}
>   {...dragHandlers}
> >
>   <div
>     className="handle cursor-move absolute"
>     id={item.id}
>     datatype={item.item_type}
>     style={{ left: `${item.x * 100}%`, top: `${item.y * 100}%` }}
>   >
>     <div className="item item--media text-serif text-styled-500">
>       {item.content}
>     </div>
>   </div>
> </Draggable>;
> ```

Did you find a way? I tried implementing a condition to implement the dragging percentage based on the parent element's width compared to the current element width, but It did not work on iPhone.

PR #64, opened 2015-06-14
Adding the option for percentag based dragging

AndrewRayCode:
This is an initial attempt at adding percentage based offsets to address #63.

One of the odd things in this plugin is the ability to cancel the drag events by returning false from onDrag. I don't think this is healthy for the API. This change initially ignore the return value in the the drag callback for code convenience.

If this works for you I can clean up the code a bit, make sure it matches style, and update the documentation.

### DIFF
--- a/lib/draggable.js
+++ b/lib/draggable.js
@@ -15,6 +15,10 @@ function createUIEvent(draggable) {
   var state = draggable._pendingState || draggable.state;
   return {
     node: draggable.getDOMNode(),
+    percentage: {
+      top: state.percentY,
+      left: state.percentX
+    },
     position: {
       top: state.clientY,
       left: state.clientX
@@ -463,7 +467,9 @@ module.exports = React.createClass({
      * A workaround option which can be passed if onMouseDown needs to be accessed,
      * since it'll always be blocked (due to that there's internal use of onMouseDown)
      */
-    onMouseDown: React.PropTypes.func
+    onMouseDown: React.PropTypes.func,
+
+    percentage: React.PropTypes.bool
   },
 
   componentWillReceiveProps: function(newProps) {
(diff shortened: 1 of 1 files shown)


supermario:
Coincidentally was after this functionality myself this evening and stumbled across your PR! :smile: 

Have tested this locally, seems to work ok with a few issues (that you're possibly already aware of):
- You'll probably want to initialize `percentX` and `percentY` in `getInitialState` to preserve the `start=` method's functionality in percent mode:

``` diff
--- a/lib/draggable.js
@@ -515,7 +514,8 @@ module.exports = React.createClass({
       offsetX: 0, offsetY: 0,

       // Current transform x and y.
-      clientX: props.start.x, clientY: props.start.y
+      clientX: props.start.x, clientY: props.start.y,
+      percentX: props.start.x, percentY: props.start.y
     };
   },
```
- The element's offset to the dragpoint isn't being correctly determined in your positioning calculation, so when you click to drag an element, it snaps the elements top left corner (`0,0`) to match your cursor, like so:

drag-snap

Had a bash at trying to fix that but couldn't wrap my head around the right numbers and ran out of time.

STRML:
I like this idea, but it will be much easier to implement on top of `<DraggableCore>`. Expect the refactor to hit in the next few days.

https://github.com/mzabriskie/react-draggable/tree/coreRefactor

STRML:
This is a TODO for me now that `v1.0.0` is out.

STRML:
Closing this PR as stale. This PR has been open since 2015 (almost 10 years) and the codebase has undergone significant changes since then. If percentage-based dragging is still desired, please open a fresh PR against the current codebase. Thank you for the contribution!

PR #68, opened 2015-06-17
Add multitouch support

brianfay:
Hi, these changes should enable multi-touch support, so that elements can be individually dragged at the same time, when using a multi-touch device.

I've tested this successfully on Chrome for Android. On Firefox for Android, scroll and zoom aren't being prevented (a separate issue?) I unfortunately do not have an iOS device to test with.

This adds a few conditional statements in the event handlers, but it didn't seem to impact performance when testing on my laptop. The original tests all pass, and everything in the demo is functioning as I would expect it to (The "Active Drags" counter is fun to watch, now)

I would like to add unit tests for this, but I'm not sure how to simulate this situation - everything depends on how the state of each component changes.

There might be a more performance-savvy way to do this by having one set of touch event listeners per app, but that would take a significant refactoring and might not be in line with the philosophy of React.

Please let me know what you think!

### DIFF
--- a/lib/draggable.js
+++ b/lib/draggable.js
@@ -516,7 +516,7 @@ module.exports = React.createClass({
   handleDragStart: function (e) {
     // Set touch identifier in component state if this is a touch event
     if(e.targetTouches){
-      this.setState({identifier: e.targetTouches[0].identifier});
+      this.setState({touchIdentifier: e.targetTouches[0].identifier});
     }
     
     // Make it possible to attach event handlers on top of this one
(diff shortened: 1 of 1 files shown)


STRML:
Hey, this is pretty nice. Thanks for the contribution.

I agree that this is the right way to do it. Any other approach essentially relies on globals.

brianfay:
Thanks! If we were using ES6 classes instead of React.createClass, it might make sense to have the event listeners be static class variables, enforcing one instance of each for the whole class. The single listener could be added with the first drag from any component, and removed with a drop as long as no other component is dragging... but determining that would also probably require global state.

I guess in the worst case, one ten-fingered person will be dragging ten items at once. There will be ten listeners that each fire the handleDrag ten times. That's definitely not ideal, but... at least most people don't have more than ten fingers!

PR #74, opened 2015-07-23
Changing translate(x, y) to translate3d(x, y, z)

connorjburton:
Adds hardware acceleration to the dragging. Should increasing performance hugely.

### DIFF
--- a/lib/draggable.js
+++ b/lib/draggable.js
@@ -211,14 +211,13 @@ function createCSSTransform(style) {
   // Replace unitless items with px
   var x = style.x + 'px';
   var y = style.y + 'px';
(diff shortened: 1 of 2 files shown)


STRML:
Could you test that it actually does speed it up? AFAIK this hack may not be necessary on modern browsers and could actually cause unwanted hardware acceleration issues.

A screenshot of the Timeline view from Chrome or FF would be great.

connorjburton:
Have done a test on what I can so far in our App and it seems negligible - however I will keep testing as I go along and add more features/elements to our drag canvas and see how it performs on browsers and iPads.

From experience is does tend to speed things up when there is lots of element moving, however may not be necessary in this case.

Happy to close this and open another when I have more proof if needed.

PR #75, opened 2015-07-23
Add calculate rotated bounds plus extends transform by child styles.

rroslaniec:


STRML:
Could you add some notes on what this accomplishes?

rroslaniec:
Hello, sure. Its fixing problem if you have draggable element (like text) with rotate css transform with bounds to his parent.

STRML:
Thanks for the contribution, but I consider this out of scope for Draggable - this is a lot of code to maintain for something that is truly niche. If you want to rotate your text, simply wrap the text in a `<span>` so the rotation doesn't mess up the transform.

If that doesn't fix your problem, feel free to build on top of `<DraggableCore>` with your custom transform.

PR #76, opened 2015-07-25
fix x/y position when axis limit is enabled

meikidd:
when axis="x" (or axis="y") is enabled, the start y (or x) position becomes 0, correct value should be this.props.start.x (or this.props.start.y)


STRML:
Thanks.

PR #77, opened 2015-08-04
Updating README to reflect child element's required API

jmuerle:
Draggable currently requires the style, className, onMouseDown, onMouseUp, onTouchStart, and onTouchEnd props to be handled by the child element. This change makes the requirement explicit in the README so that users will hopefully not be confused in the future.


STRML:
Thanks.

PR #86, opened 2015-09-27
Draggables inside scrollable parent (not document.body) - bug fix

dkozar:
Although the current version of react-draggable respects document.body scroll, there's a problem when draggables are inside of another scrollable container.

This fork makes draggable listen to a specified ancestor, rather than document.body (but it defaults to document.body if ancestor not specified).

Example: https://www.youtube.com/watch?v=FZ-L5x4E02k

Cheers!

Danko

scrollable1
scrollable2


dkozar:
Closed this one because created a new commit and pull request: https://github.com/mzabriskie/react-draggable/pull/87

PR #90, opened 2015-10-07
Adds prop 'isSVG' (and respective functions) which allows for proper transform on SVG elements.

martinRoss:
Updated specs to include new tests for this functionality. Was hoping to automatically detect if a node is SVGElement but does not look like there is a good way to do so.

I've been having issues with my laptop - please give the tests a run before merging!


mzabriskie:
Can't this just be detected? Why do we need to explicitly say so?

martinRoss:
@mzabriskie You're right, we should be able to detect; I simply was unsuccessful in doing so. Unfortunately, my laptop's hard drive just kicked it; it's going to be a while before I can get back to this.

Below was my approach so far that turned out to be unsuccessful. Any ideas? One flaw I see is that if an element gets passed in (say, a `<g />` ), `instanceof` might only work when rendering into an SVG container. Bit of a chicken or egg problem; we need to know what type of element it is before we render it into that test container...

https://gist.github.com/martinRoss/432affa8b19732e9504e

martinRoss:
Took a new approach and found a way to detect. Will update tests and submit a new pull request in a few days.

PR #91, opened 2015-10-09
pre might be null or undefined 

jamwise:
This PR fixes the fringe case where pre is `undefined` or `null`, in one of those cases calling `.slice()` will throw an error. This happened when using this library in tests that make use of jsdom.


rm-hull:
+1 - causing me pain too - last working version against JSDOM was 0.8.1

STRML:
Thanks for this, and @rm-hull for getting it back in my inbox - easy merge.

PR #93, opened 2015-10-10
Example of broken on start change

AndrewRayCode:


STRML:
`moveOnStartChange` is now gone, use `<DraggableCore>` if you need more control.

PR #95, opened 2015-10-11
Fixing README to update code problem

AndrewRayCode:


STRML:
That's fine - I'll merge this for now, but the feature is going to die as part of the rewrite.

PR #97, opened 2015-10-11
Add install static-server to readme

TSMMark:
`static-server` command is a requirement of build-watch so it makes sense to mention it in readme.


STRML:
The script should instead be amended to reach out to
node_modules/.bin/static-server

STRML:
Fixed in https://github.com/mzabriskie/react-draggable/commit/101d897cb435a54ff51ec3e878ed01c73024ed41

PR #99, opened 2015-10-15
Adds detection of SVGElements and uses appropriate transform attribute.

martinRoss:
IE does not support transforming SVGElements via the style attribute and requires the change to be via the transform attribute. Other browsers also respect these attributes so it should not break elsewhere.

Added tests for both SVG detection and proper use of SVG transforms. Let me know if you'd like to see anything else!


psaia:
Awesome!

STRML:
Thanks! This has some conflicts, most probably in the dist files, so I'll need to pull it and commit separately.

STRML:
@martinRoss Could you rebase on master?

martinRoss:
Sure thing, I'll rebase and put in a new pull request. You might still have
some issue with the dist files though.

STRML:
Doesn't need to be a new PR, you could just rework this one. Just skip the dist files entirely or regenerate them in the last commit.

martinRoss:
Ah, whoops, still new to PRs--created a new one. Regenerated the dist files
in the last commit.

PR #100, opened 2015-10-15
Adds detection of SVGElements and uses proper transform attribute. Rebased back onto upstream master.

martinRoss:
Dist is still going to have conflicts, you might just want to accept or delete and rebuild.

PR #101, opened 2015-10-16
fixed isElementSVG state property issue so it does not get overwritte…

psaia:
Hey guys, this is a follow up bug fix to @martinRoss 's fix to [SVG elements in IE](https://github.com/mzabriskie/react-draggable/pull/100).

There were two issues:
1. `isElementSVG` is being defined in `getInitialState()` . This was causing the state to be overwritten (and set back to false) on `componentWillReceiveProps()`. I simply added a check to see if the state already exists, and if it does, it won't re-set to false.
2. Down in `render()` we should apply the CSS style for transform despite whether or not the element is a SVG. Modern browsers do not respect the real `transform` attribute - so essentially the problem was reversed and fixed in IE but broken in Chrome, FF, ect.

Tested locally and it's working great. Thanks!

### DIFF
--- a/lib/draggable.js
+++ b/lib/draggable.js
@@ -504,12 +504,13 @@ module.exports = React.createClass({
       onStart: emptyFunction,
       onDrag: emptyFunction,
       onStop: emptyFunction,
-      onMouseDown: emptyFunction,
+      onMouseDown: emptyFunction
     };
   },
 
   getInitialState: function (props) {
     // Handle call from CWRP
+    var currentState = this.state;
     props = props || this.props;
     return {
       // Whether or not we are currently dragging.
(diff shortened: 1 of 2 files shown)


martinRoss:
Thanks for the fix @petesaia. I haven't noticed any modern browsers ignoring the real `transform` attribute--out of curiosity, in which browsers does this break? No issues in the most recent Chrome, FF, for me. Are there any downsides to having both `CSS transform` and `transform` at the same time?

psaia:
I noticed that in Chrome and Firefox it was not respecting the `transform` attribute. It was just stuck at 0. Though it sounds like support for the `transform` attribute is inconsistent. After a lot of playing around, i'm not seeing any side effects in applying both.

psaia:
FWIW, I'm applying it on a `<g>` element.

martinRoss:
Very strange. I'm applying it to a <g> element as well and it's working
perfectly on Chrome 46 and FF 43. On Linux but doubt that matters for this
case.

martinRoss:
http://www.w3.org/TR/SVG/coords.html#TransformAttribute

psaia:
@martinRoss You're absolutely right. I had a style set in CSS which was overwriting the `transform`. I will commit, squash, and push.

psaia:
Done!

martinRoss:
Sweet!

STRML:
Thanks!

PR #111, opened 2015-11-11
:memo: the first argument is still event, the structure was just for second argument

basarat:
Great project BTW. Buttery smooth :rose:

resize


STRML:
Thanks!

PR #114, opened 2015-11-23
Fix server-side rendering

liseki:
Handle pseudo-browser environments when rendering server-side where `window` may be defined as the global context like in [react-rails](https://github.com/reactjs/react-rails/blob/024cdbf64c03038503788a783b393e13cdaa22c7/lib/react/server_rendering/exec_js_renderer.rb#L34-L36)

### DIFF
--- a/lib/utils/getPrefix.es6
+++ b/lib/utils/getPrefix.es6
@@ -1,5 +1,9 @@
 export default (function() {
-  if (typeof window === 'undefined') return '';
+  // Checking specifically for 'window.document' is for pseudo-browser server-side
+  // environments that define 'window' as the global context.
+  // E.g. React-rails (see https://github.com/reactjs/react-rails/pull/84)
+  if (typeof window === 'undefined' || typeof window.document === 'undefined') return '';
+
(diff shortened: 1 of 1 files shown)


STRML:
Wow - this seems like a bug in `react-rails`. It shouldn't be mocking the window object for this exact reason.

If it's going to mock `window` and not provide any of its methods, what is the point?

liseki:
Sorry for initially breaking a test, it seems fine now.

@STRML I absolutely agree with you on that. Looking at the [original patch](https://github.com/reactjs/react-rails/pull/84) that issue was raised but never addressed albeit after the commit. Looking at the discussion this was partly done to make it play happy with browserify. Considering all that is this small patch acceptable or will I have to monkey patch react-rails?

STRML:
I suppose it is fine if it fixes your use case, but I very much disagree with what react-rails did here. Could you add a comment (so it doesn't get deleted later)? I'll merge it then.

liseki:
Thanks @STRML, I think that should be ready to merge now!

STRML:
Thanks!

PR #117, opened 2015-12-11
update draggable position after update props

dramoturg:


STRML:
This is the reason why `<Draggable>` was split into `<Draggable>` and `<DraggableCore>`, use the core module if you need direct control.

dramoturg:
That is, I need to create a component that will use Draggable ~DraggableCore~. And it defines componentDidMount method?

Issue #122, opened 2016-01-02
Running react-draggable in system.js / jspm


erfangc:
Using jspm / systemjs, I've installed `react`, `react-draggable`

``` html
<!doctype html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Document</title>
    <script src="jspm_packages/system.js"></script>
    <script src="config.js"></script>
    <script>
        System.import("./DnD");
    </script>
</head>
<body>
<div id="content">

</div>
<style>
    #content {
        width: 200px;
        height:200px;
        border: 1px solid #000;
    }
</style>
</body>
</html>
```

Entry File `DnD.js`:

``` js
import * as React from 'react';
import * as ReactDOM from 'react-dom';
import * as Draggable from 'react-draggable';

export class DragTest extends React.Component {

    constructor(props) {
        super(props);
    }

    onStart() {
        console.log("drag started");
    }

    onStop() {
        console.log("drag stopped");
    }

    render() {
        var drags = {onStart: this.onStart, onStop: this.onStop};
        return (
            <Draggable {...drags}>
                <div>
                    <h1>
                        Hello World!
                    </h1>
                </div>
            </Draggable>
        );
    }
}
debugger;
ReactDOM.render(<DragTest />, document.getElementById("content"));
```

With the following transpiler options

``` json
transpiler: "babel",
  babelOptions: {
    "blacklist": [],
    "optional": [
      "runtime",
      "optimisation.modules.system"
    ]
  }
```

The error in DEV console (Chrome 47): 
React.createElement: type should not be null, undefined, boolean, or number ... 

Inspecting the names at the `debugger;` line, we see that `DragTest`, my custom test component, is a constructor `function`, which I expected it to be. 

However, `Draggable` in this case looks like:

``` json
{displayName: "Draggable", propTypes: Object, defaultProps: Object ... }
```

Is this a known issue / or am I doing something wrong? Looking at the code, `./lib/*.es6` are all written in ES6 with `import` style module loading syntax, whereas the `index.js` file uses a node commonjs style `module.exports = require` style.

erfangc:
Having played around with `system.js` some more (and recalling similar behavior I saw with other libraries under `system.js`), it appears the Draggable component can be accessed under `.default` of the imported value

i.e. 

``` js
import * as ReactDraggable from 'react-draggable';
var Draggable = ReactDraggable.default;
```

[...]

PR #124, opened 2016-01-02
Update README.md

erfangc:
#122 clarifying how to import the component using a ES6 loader instead of browserify

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,20 @@ This bundle is also what is loaded when installing from npm. It expects external
 If you want a UMD version of the latest `master` revision, you can generate it yourself from master by cloning this
 repository and running `$ make`. This will create umd dist files in the `dist/` folder.
 
+### Installing with jspm and systemjs ES6 module loader
+
+```bash
+$ jspm install npm:react-draggable
+```
+
+To access the `Draggable` component in your own code, you refer to the `.default` property of the imported reference.
+
+```js
+import * as ReactDraggable from 'react-draggable';
+var Draggable = ReactDraggable.default;
+ReactDOM.render(<Draggable />, ...);
+```
+
 ## Draggable
 
 A `<Draggable>` element wraps an existing element and extends it with new event handlers and styles.


STRML:
This seems like a bug in jspm/SystemJS, not in React-Draggable. This community will continue to run into this issue again and again until they're able to handle babel-transpiled modules.

I'm not sure if I really understand the issue. Is jspm using dist/, not lib/?

erfangc:
jspm is using lib/ it appears, but still need the .default to work

STRML:
Ah, this is an older project so it is laid out slightly differently than I usually do now. 

Usually my layout is:

```
src/  <-- es6 files
lib/  <-- es5 files compiled with Babel for use with Node
dist/ <-- bundled files for use in browsers
```

But this project skips a step, using the `dist/` folder for both Node and Browser installs. This should be fine, though, and works just fine in Node.

I'm still not sure what JSPM is doing here to make it fail. See [this main export line](https://github.com/mzabriskie/react-draggable/blob/master/dist/react-draggable.js#L59), which requires from [this export](https://github.com/mzabriskie/react-draggable/blob/master/dist/react-draggable.js#L363). Notice the line after, where `module.exports` is assigned to `exports['default']`. So there shouldn't be an issue.

Some clues, perhaps: see [this issue](https://github.com/systemjs/systemjs/issues/304) talking about why SystemJS has decided not to detect `default` and require it.

erfangc:
This is definitely a limitation of SystemJS import, one that may or may not have any relevance as module become more universally more ES6 like. It is not a big deal, just annoying and time consuming to debug the first couple of times.

STRML:
Sure. Well, hell, doesn't hurt anything to put it in the README. Thanks @erfangc.

STRML:
Actually, now that I think of it, wouldn't it make sense to just:

``` js
import ReactDraggable from 'react-draggable';
```

Does that not work? It makes sense that you'd have to access `.default` if you did an `import *`.

erfangc:
Doesn't the react-draggable module export `DraggableCore` as well? I thought `import *` would seem cleaner and quicker ... but I suppose `import {ReactDraggable, DraggableCore}` works too? (haven't tested, will test and see if that works)

STRML:
Would be:

``` js
import {default as Draggable, DraggableCore} from 'react-draggable';
```

I may back this out, this is just how ES6 loaders work, regardless of whether or not you're using SystemJS or Babel.

erfangc:
make sense, thanks for that! I guess I didn't read the full spec.

PR #125, opened 2016-01-03
Skip handleDrag if disabled

bokuweb:
Skip the `handleDrag` When disabled, because, If it becomes disabled at the drag start event, it would be dragged.


STRML:
This seems like an odd thing to do. Why not just return false from `onStart`?

bokuweb:
@STRML thanks
If become disable while dragging, I want to skip the drag processing

STRML:
You can do this now by returning false from `onDrag`.

PR #127, opened 2016-01-05
Fixed updating of start position via parent using componentWillReceiveProps

josdeweger-zz:
…

### DIFF
--- a/lib/Draggable.es6
+++ b/lib/Draggable.es6
@@ -138,6 +138,13 @@ export default class Draggable extends DraggableCore {
     }
   }
 
+  componentWillReceiveProps(nextProps) {
+    this.setState({
+      clientX: nextProps.start.x,
+      clientY: nextProps.start.y
+    });
+  }
+
   onDragStart = (e, coreEvent) => {
     log('Draggable: onDragStart: %j', coreEvent.position);
 
--- a/specs/draggable.spec.jsx
+++ b/specs/draggable.spec.jsx
@@ -190,6 +190,50 @@ describe('react-draggable', function () {
       expect(style.indexOf('transform: translate(100px, 100px);')).not.toEqual(-1);
     });
 
+    it('should translate the initial start x and y position to a transform style', function () {
+      drag = TestUtils.renderIntoDocument(
+        <Draggable start={{x: 50, y: 50}}>
+          <div />
+        </Draggable>
+      );
+
+      var node = ReactDOM.findDOMNode(drag);
+      var style = node.getAttribute('style');
+
+      expect(style.indexOf('transform:translate(50px,50px);')).not.toEqual(-1);
+    });
+
+    it('should update start x and y position to a transform style', function () {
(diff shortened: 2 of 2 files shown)


lib/Draggable.es6
```diff
@@ -138,6 +138,13 @@ export default class Draggable extends DraggableCore {
     }
   }
 
+  componentWillReceiveProps(nextProps) {
```
STRML, line 141: This would cause the draggable to jump if you changed any property at all, or even if a render cascaded down to this element.

josdeweger-zz, line 141: Hm good point. Any suggestions on how to solve this issue?

I have the following setup:

Component `<Lane>` contains multiple `<Batch>`

```
export default class Lane extends React.Component {
  [...]

  constructor(props) {
    super(props);

    this.state = {
      visibleLaneWidth: 0
    };
  }

  componentDidMount() {
    this.setState({
      visibleLaneWidth: this.refs.laneContent.offsetWidth
    });
  }

  render() {
    [...]
    const { visibleLaneWidth } = this.state;

    return (
      [...]
      <div ref="laneContent">
          {batches.map((batch) => {
            return (
              <Batch key={batch.id}
                    id={batch.id}
                    title={batch.title}
                    status={batch.status}
                    period={batch.period}
                    visiblePeriod={visiblePeriod}
                    visibleLaneWidth={visibleLaneWidth} />
            );
          })}
      </div>
    );
  }
}
```

Component `<Batch>` contains a `<Draggable>`:

```
export default class Batch extends React.Component {
   calculatePositionX() {
     var visiblePeriodStartDate = this.props.visiblePeriod.start;
     var visibleNrOfDays = this.props.visiblePeriod.diff('days');
     var diffStartDateBatchStartDateVisiblePeriod = this.props.period.start.diff(visiblePeriodStartDate, 'days');
     var xPos = (diffStartDateBatchStartDateVisiblePeriod / visibleNrOfDays) * this.props.visibleLaneWidth;

     return xPos;
  }

  render() {
      [...]
      const posX = this.calculatePositionX();
      const posY = 0;
      [...]

      return (
         <span>
           <Draggable
             axis="x"
             start={{x: posX, y: posY}}
             moveOnStartChange={true} grid={grid}>
             <div />
           </Draggable>
         </span>
       );
    }
}
```

The method calculatePositionX depends on the prop visibleLaneWidth, which is at first 0, so the initial outcome of the method is also 0. As soon as the visibleLaneWidth changes the method gets called again and returns the correct number, however, the Draggable is not updated and stays at x position 0. You can see this failing by executing the test in this PR.

STRML, line 141: Moveonstartchange has been removed.

You can either use DraggableCore, or attach a ref to the Draggable el and
call setState on it directly.

kis, line 141: @STRML could you please explain how to use DraggableCore, or make a ref to the Draggable el?

PR #132, opened 2016-01-20
Fixed: mouse not working properly on touch screen laptops

dkozar:
The "touch & mouse" bug is still alive and well (I'm aware there have been few attempts to squash it).

After you use touch, you can't use mouse anymore - this could be reproduced in your demo.

You could emulate touch screen laptop with Chrome. In Chrome inspector, click the "mobile" icon:

Image 1

Your browser window will get a border and additional options. Choose the dropdown option "laptop with touch":

Image 2

Now touch one of the draggables with a mouse (you'll notice that when touch emulation is ON, the mouse pointer turns into a circle).

After that, turn OFF the touch emulation, by unchecking this checkbox in the inspector:

Image 3

At this point you won't be able to drag with a mouse anymore.

That's because of this return statement, and also the absence of the line setting the input back to mouse:

Image 4

You could easily track this by console-logging the handleDragStart method:

handleDragStart = (e) => {
    console.log('handleDragStart', e);
    // ....

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 (MIT License)
 
-Copyright (c) 2014 Matt Zabriskie. All rights reserved.
+Copyright (c) 2014-2016 Matt Zabriskie. All rights reserved.
  
 Permission is hereby granted, free of charge, to any person obtaining a
 copy of this software and associated documentation files (the "Software"), 


STRML:
Thanks for the fix and this detailed writeup.

Could you please rebase this to a single commit? I'll merge it shortly thereafter.

STRML:
No need to close, please just rebase and force push.

dkozar:
Sorry, I just created another pull request with the same diff.

dkozar:
Could you please take a look at another one (https://github.com/mzabriskie/react-draggable/pull/135) or allow me to force push to your master?

PR #135, opened 2016-01-22
Fixed: mouse not working properly on touch screen laptops

dkozar:
Rebased as requested and ready for merging with a repo.

### DIFF
--- a/lib/DraggableCore.es6
+++ b/lib/DraggableCore.es6
@@ -392,6 +392,18 @@ export default class DraggableCore extends React.Component {
     this.props.onDrag(e, coreEvent);
   };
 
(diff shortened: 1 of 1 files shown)


STRML:
[This article](http://ariatemplates.com/blog/2014/05/ghost-clicks-in-mobile-browsers/) has some more information on how one might reliably detect/prevent ghost clicks.

Would you be willing to put a `preventDefault()` in on the `touchend` event? I don't think we have to go whole-hog on screwing with timers - but I don't have a hybrid mouse/touchscreen device at home so I don't have a good way to test if it helps or not.

dkozar:
I'll take a look.

dkozar:
Rebased to last version.

PR #138, opened 2016-02-08
Updates year in changelog for last entry to 2016

houshuang:


STRML:
This is fixed; thanks.

PR #145, opened 2016-03-21
Fixed constructor bug in IE10

codeflyer:
The props are not available in the constructor in IE10, adding a check on this.props before using it helps to prevent an error.

### DIFF
--- a/lib/Draggable.es6
+++ b/lib/Draggable.es6
@@ -148,7 +148,7 @@ export default class Draggable extends React.Component {
     dragged: false,
 
     // Current transform x and y.
-    clientX: this.props.start.x, clientY: this.props.start.y,
+    clientX: this.props ? this.props.start.x : 0, clientY: this.props ? this.props.start.y : 0,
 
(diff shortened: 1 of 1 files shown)


STRML:
Is this a Babel bug?

STRML:
Perhaps `proto-to-assign` will fix this - see https://babeljs.algolia.com/docs/advanced/caveats/#internet-explorer

codeflyer:
Thanks for the quick answer, I'm checking, I'll let you know ASAP

codeflyer:
I've tried to add the transform but i still have the error:
"unable to get property 'start' of undefined or null reference"

[
 'transform-es2015-block-scoping',
 ['transform-es2015-classes', {loose: true}],
 'transform-proto-to-assign'
]

With this solution the call of a constructor works and if I try to log the props in the constructor are finally available (without this transform thy are not available https://github.com/babel/babelify/issues/133)

But in the code you don't use a contructor, but static calls. 

...I'll continue on my search

STRML:
This was fixed in #158.

PR #149, opened 2016-04-05
Fix for touchscreen laptops: input could be switched between mouse an…

dkozar:


STRML:
This makes sense - thanks.

PR #153, opened 2016-04-20
Only apply user-select hack if drag is started

davidstubbs:
Previously the user-select hack is applied whenever mouseDown on an
element matching the selector, with the right kind of click.

If the onStart method returns false then the drag is aborted, but
the user-select hack is still applied. Because the drag is not running,
the hack is not unapplied on mouseUp. This leads to a whole series of
";user-select: none;; user-select: none;; ..." appearing on the body tag
and the inability to select text on the page until a valid drag happens.

This commit moves the hack application to after the last short circuit
to ensure dragging is actually happening before disabling selection.

I've added a test case and also a box on the example page.

### DIFF
--- a/example/index.html
+++ b/example/index.html
@@ -135,6 +135,9 @@
           <Draggable axis="y" {...dragHandlers}>
             <div className="box cursor-y">I can only be dragged vertically</div>
           </Draggable>
+          <Draggable onStart={() => false}>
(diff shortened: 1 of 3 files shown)


STRML:
Nice work! I've seen this bug before but didn't have the time to track it down. Solid PR, and thanks for not committing built files.

PR #156, opened 2016-05-17
[Enhancement] Support nodes within other frames

Craga89:
Added a `getOwnerDocument` utility that resolves the correct `document` based on the `DraggableCore`s associated DOMElement. This enables usage of `react-draggable` on nodes that are rendered within an `<iframe />`.

### DIFF
--- a/lib/DraggableCore.es6
+++ b/lib/DraggableCore.es6
@@ -1,6 +1,7 @@
 // @flow
 import React, {PropTypes} from 'react';
-import {matchesSelector, addEvent, removeEvent, addUserSelectStyles,
+import { findDOMNode } from 'react-dom';
+import {getOwnerDocument, matchesSelector, addEvent, removeEvent, addUserSelectStyles,
         removeUserSelectStyles, styleHacks} from './utils/domFns';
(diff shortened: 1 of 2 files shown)


STRML:
Hey, nicely done!

Could you add a quick test case for it?

Craga89:
Hmmm, I'm not too up-to-date with the `TestUtils`! Currently using `react-frame-component` to render into an Iframe, so we'd probably need to do that, but since it uses the `unstable_renderSubtreeIntoContainer` ReactDOM utility not sure if it would work within the tests?

Craga89:
Actually, maybe I just need to test that `getOwnerDocument` works as expected? That should be easy enough.

STRML:
Yeah - rendering iframed components is a little crazy IMO and breaks all sorts of assumptions. To test it properly I don't mind installing react-frame-component as a devDep.

Craga89:
:+1: I'll give it a shot when I get some time

STRML:
Closing this in favor of #177 which also handles `window.getComputedStyle()`, which has problems with iframes.

STRML:
Thanks anyhow for the contribution and I apologize that it got left here for this long.

PR #158, opened 2016-05-19
Use explicit constructor in Draggable to fix IE10 null reference

davidstubbs:
There's a problem in IE10 with ES2015 classes and the super class.
Essentially the prototype isn't set up correctly which leads to
this.props being null when setting up position in the construction.

However, props are passed to the class constructor explicitly by React,
so we can use them from there.

Tested by running the examples in IE 10 on Win 7.

This is basically the same problem as addressed in #145, but fixed to work rather than just not blow up (and for version 2 props).

### DIFF
--- a/lib/Draggable.es6
+++ b/lib/Draggable.es6
@@ -18,6 +18,11 @@ type DraggableState = {
   isElementSVG: boolean
 };
 
+type ConstructorProps = {
+  position: { x: number, y: number },
+  defaultPosition: { x: number, y: number }
+}
+
 //
 // Define <Draggable>
 //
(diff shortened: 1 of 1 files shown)


STRML:
So it appears class property assignments are broken in Babel with IE < 11?

davidstubbs:
No, it's more that accessing properties placed on the object by the super constructor is broken.

`this.props` works no better in the explicit constructor than in a class property.

For example, writing this for Draggable:

``` javascript
 constructor(props) {
  super(props);
  console.log(this.props);
 }
```

On chrome: `Object { zIndex: 100, children: Object ... }`
On IE10: `undefined`

As I understand it there's a basic incompatibility between IE <= 10 and babel ES2015 classes because you can't set the prototype properly. There's a bunch of work arounds but they all have edge cases and side effects.

Using the explicit constructor just gives us independent access to the props as part of the published React API.

STRML:
:+1: Thanks for the further information. It's a shame this bug exists, and I appreciate you taking the time to fix it.

STRML:
Published in `v2.0.1`.

Issue #118, opened 2015-12-15
Get Control Position on Mobile


ghost:
Not really a major issue if you're just dragging elements and not recording the positions anywhere... but we were having trouble returning the position of a dragged element on mobile devices and traced it back to getControlPosition()

We fixed this by changing:
var position = e.targetTouches && e.targetTouches[0] || e;
to
var position = e.changedTouches && e.changedTouches[0] || e;

targetTouches was returning empty for us in iOS and Android.

STRML:
Fixed in `2.0.2`.

PR #159, opened 2016-05-19
Fix for 'cannot access clientX on undefined' exception.

bryanlarsen:
To reproduce, access a DraggableCore element using touch emulation in Chrome.   The exception will be thrown on drag stop.

This will leave the element active, so if you drag another element, both elements end up being dragged together.

I'm not sure if this is the correct fix, I may be treating symptoms rather than root cause, but it appears to work.


STRML:
This may be related to #118 - looking into it.

STRML:
`2.0.2`

PR #166, opened 2016-06-17
Fix for latest iPad scrolling instead of dragging vertically.

JaneCoder:
This tweak stops page scrolling while dragging, which was keeping ipad and ipad pro from allowing drag of the draggable elements. I added an eventListener while dragging, and removed it on touch end.

### DIFF
--- a/lib/DraggableCore.es6
+++ b/lib/DraggableCore.es6
@@ -263,11 +263,10 @@ export default class DraggableCore extends React.Component {
         this.handleDragStop(new MouseEvent('mouseup'));
       } catch (err) {
         // Old browsers
-        const event = document.createEvent('MouseEvents');
+        const event = ((document.createEvent('MouseEvents'): any): MouseEvent);
         // I see why this insanity was deprecated
         // $FlowIgnore
         event.initMouseEvent('mouseup', true, true, window, 0, 0, 0, 0, 0, false, false, false, false, 0, null);
-        // $FlowIgnore
         this.handleDragStop(event);
       }
       return;


lib/DraggableCore.es6
```diff
@@ -236,6 +239,11 @@ export default class DraggableCore extends React.Component {
   };
 
   handleDrag: EventHandler<MouseEvent> = (e) => {
+    // Stop the touch device from scrolling while user is dragging.
+    if(this.state.scrolling == true && this.state.dragging) {
```
STRML: This could simply be `if (this.state.scrolling && this.state.dragging)` - please be aware of the existing style, spacing, etc.

lib/DraggableCore.es6
```diff
@@ -236,6 +239,11 @@ export default class DraggableCore extends React.Component {
...
+      document.addEventListener('touchmove', this.removeScroll)
+      this.state.scrolling = false;
```
STRML: You have to use `this.setState({})` in React, don't assign to state directly.

lib/DraggableCore.es6
```diff
@@ -294,6 +302,7 @@ export default class DraggableCore extends React.Component {
     // Reset the el.
     this.setState({
       dragging: false,
+      scrolling: true,
```
STRML: This is going to cause us to always assign a 'touchmove' handler if the draggable has been dragged more than once.

lib/DraggableCore.es6
```diff
@@ -331,6 +345,10 @@ export default class DraggableCore extends React.Component {
     // We're on a touch device now, so change the event handlers
     dragEventFor = eventsFor.touch;
 
+    // Remove the event listener that stopped document scrolling
+    document.removeEventListener('touchmove', this.removeScroll, false);
```
STRML: In fact, why don't we just add/remove this listener `onTouchStart` outright? So no `state.scrolling` is required at all.

Also be sure to add an extra `document.removeEventListener` on unmount in case an unmount happens while dragging.

STRML:
Thanks for the contribution - some notes for you - also please exclude any built files from PRs.

lib/DraggableCore.es6
JaneCoder: Thank you so much! I was not wanting it to redraw the page while they were dragging. 
Any advice on a good learning resource for someone new to front end dev?

lib/DraggableCore.es6
JaneCoder: Ah, I was trying to set back to true after making it false on end touch.

lib/DraggableCore.es6
JaneCoder: But then when the user is touching the device in order to scroll, won't that stop scrolling when they want to scroll?
Would I check for dragging state to fix that?

STRML: You mean to scroll a div inside the element itself? Because the touch handler is only on the draggable element.

JaneCoder: Yes, actually some of my draggables have scrollable divs inside. I'll  test that.

JaneCoder: OK, it works without disturbing internal scrolling for us. I did use the draggableCancel and draggableHandle properties though. 

We will do some testing without those to see if it could be an issue for others. Thanks so much for your great product.

PR #167, opened 2016-06-17
fix for ipad and ipad pro vertical drag stopped by page scrolling

JaneCoder:
Trying this again hopefully without any built files. This patch adds an event listener to the document to stop page scroll triggering instead of vertical dragging.


STRML:
We're close, but not quite there. Did this work in your testing?

lib/DraggableCore.es6
```diff
@@ -175,6 +175,7 @@ export default class DraggableCore extends React.Component {
     removeEvent(document, eventsFor.touch.move, this.handleDrag);
     removeEvent(document, eventsFor.mouse.stop, this.handleDragStop);
     removeEvent(document, eventsFor.touch.stop, this.handleDragStop);
+    document.removeEventListener('touchmove', this.removeScroll);
```
STRML: Please use the `removeEvent` helper for IE compat.

lib/DraggableCore.es6
```diff
@@ -319,18 +320,31 @@ export default class DraggableCore extends React.Component {
...
+  };
+
   // Same as onMouseDown (start drag), but now consider this a touch device.
   onTouchStart: EventHandler<MouseEvent> = (e) => {
     // We're on a touch device now, so change the event handlers
     dragEventFor = eventsFor.touch;
 
+    // Stop the touch device from scrolling while user is dragging.
+    if (this.state.dragging) {
+      document.addEventListener('touchmove', this.removeScroll)
```
STRML: Please use the `addEvent` helper.

lib/DraggableCore.es6
```diff
@@ -319,18 +320,31 @@ export default class DraggableCore extends React.Component {
...
+    return false;
+  };
+
   // Same as onMouseDown (start drag), but now consider this a touch device.
   onTouchStart: EventHandler<MouseEvent> = (e) => {
     // We're on a touch device now, so change the event handlers
     dragEventFor = eventsFor.touch;
 
+    // Stop the touch device from scrolling while user is dragging.
+    if (this.state.dragging) {
```
STRML: I don't believe this will ever fire. `this.state.dragging` is set to true in `handleDragStart`.

lib/DraggableCore.es6
STRML: Also, why put this on the document, and not the draggable?

lib/DraggableCore.es6
JaneCoder: Strange testing in ipad worked. I'll test again Monday with a console out inside the if statement.

lib/DraggableCore.es6
JaneCoder: It is on the document to stop the document from scrolling on touch drag. I want the draggable to respond to touchmove, not the page.

lib/DraggableCore.es6
JaneCoder: No not firing in debug.

So I tried putting the event add in handDragStart and got rid of the dragging condition.  But then the internal scrolling of my widgets was lost. Not sure yet why.

JaneCoder:
I ended up creating an ipad_fix branch for us, and going back to the way I did it initially. Putting back the 'scrolling' state and putting the add event in handleDrag. That works for us and keeps the internal scrolling. 

I'll continue to look at this master branch and try to improve the solution, and get rid of the scrolling state, in my spare time.

lib/DraggableCore.es6
JaneCoder: Would you share a link to info on the add or remove event helpers? I'm not sure what those are.

STRML: See the top of the file, where I import the helpers. You can see the module where they're defined there. It simply helps with IE compat.

PR #168, opened 2016-06-21
defaultPosition => position

billcz:
defaultPosition => position

### DIFF
--- a/lib/Draggable.es6
+++ b/lib/Draggable.es6
@@ -121,7 +121,7 @@ export default class Draggable extends React.Component {
      *      let App = React.createClass({
      *          render: function () {
      *              return (
-     *                  <Draggable defaultPosition={{x: 25, y: 25}}>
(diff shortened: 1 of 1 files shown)


STRML:
????

nikolas:
There could have been an explanation with this PR, but it looks like it does in fact correct a typo in the documentation. Currently, the comment above Draggable.position() says defaultPosition, not position.

`````` diff
@@ -121,7 +121,7 @@ export default class Draggable extends React.Component {
       *      let App = React.createClass({
       *          render: function () {
       *              return (
 -     *                  <Draggable defaultPosition={{x: 25, y: 25}}>
 +     *                  <Draggable position={{x: 25, y: 25}}>
       *                      <div>I start with transformX: 25px and transformY: 25px;</div>
       *                  </Draggable>
       *              );
      *          }
      *      });
      * ```
      */
     position: PropTypes.shape({
       x: PropTypes.number,
       y: PropTypes.number
     }),
``````

PR #173, opened 2016-07-17
Add `offsetParent` prop for alternative offset calculations

### LINKED ISSUE #170: dragging elements on left border causes `deltaX` to jump between negative and positive values
This issue is a copy of https://github.com/STRML/react-resizable/issues/38 because it looks like the root cause is something within this library.

I have a resizable div which is `float: right`. The handler is on the left side. Everything works fine when the handler is on the right side but I'm having some troubles with the one on the right.

The following gif shows two screens. One, where the resizable is floating left and one where it's floating right. The floating left example is included to show you that it generally works.

jumpingstuff

After some debugging and seeing that it works fine if I scale the **inner** element only, I believe that the issue is caused because the origin of the element wrapping resizeable (which is top left I guess?) changes:

jumpingstuff3

In pseudocode, the react structure of the borken example is:

``` jsx
<div width="123px" float="right">
  <ReactResizable>
      <children />
  </ReactResizable>
</div>
```

The react structure of the example without jumping is:

``` jsx
<div float="right">
  <ReactResizable>
    <div width="123px">
      <children />
    </div>
  </ReactResizable>
</div>
```

In the first case, `<ReactResizable>` changes the position of the top left corner, which might have something to do with the jumping. In the second, the position of `<ReactResizable>` stays the same while the right side of the `<div>` is being changed.

I believe that the root cause for this issue is this library (react-draggable), because:

`deltaX` in [`resizeHandler`](https://github.com/STRML/react-resizable/blob/master/lib/Resizable.jsx#L135) is jumping around between negative and positive values when the issue occurs:

deltajumper

This does not happen when dragging an object on the right side, which is working fine:

delta2

[`resizeHandler`](https://github.com/STRML/react-resizable/blob/master/lib/Resizable.jsx#L135) is a `DraggableEventHandler`, so the wrong values are originating somewhere here.

aeneasr:
Allows nodes to use the body as origin instead of the parent. This is useful, when the parent's position is changing. When used, resolves #170

This issue was introduced by a398097ebcc2cbb4df5582a8a3f42f51d21745a0

please :eyes:

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,7 @@ on itself and thus must have callbacks attached to be useful.
   cancel: string,
   disabled: boolean,
   enableUserSelectHack: boolean,
+  offsetParent: HTMLElement,
   grid: [number, number],
   handle: string,
   onStart: DraggableEventHandler,
--- a/lib/DraggableCore.es6
+++ b/lib/DraggableCore.es6
@@ -67,6 +67,16 @@ export default class DraggableCore extends React.Component {
     enableUserSelectHack: PropTypes.bool,
 
     /**
+     * `offsetParent`, if set, uses the passed DOM node to compute drag offsets
+     * instead of using the parent node.
+     */
+    offsetParent: function(props, propName) {
+      if (process.browser && props[propName] && props[propName].nodeType !== 1) {
+        throw new Error('Draggable\'s offsetParent must be a DOM Node.');
+      }
+    },
+
+    /**
      * `grid` specifies the x and y that dragging should snap to.
      */
     grid: PropTypes.arrayOf(PropTypes.number),
@@ -152,6 +162,7 @@ export default class DraggableCore extends React.Component {
     cancel: null,
     disabled: false,
     enableUserSelectHack: true,
+    offsetParent: null,
     handle: null,
     grid: null,
     transform: null,
--- a/lib/utils/domFns.es6
+++ b/lib/utils/domFns.es6
@@ -2,6 +2,7 @@
 import {findInArray, isFunction, int} from './shims';
 import browserPrefix, {getPrefix, browserPrefixToStyle, browserPrefixToKey} from './getPrefix';
 
+import type DraggableCore from '../DraggableCore';
 import type {ControlPosition} from './types';
 
 let matchesSelectorFunc = '';
@@ -95,9 +96,9 @@ export function innerWidth(node: HTMLElement): number {
 }
 
 // Get from offsetParent
-export function offsetXYFromParentOf(evt: {clientX: number, clientY: number}, node: HTMLElement & {offsetParent: HTMLElement}): ControlPosition {
-  const offsetParent = node.offsetParent || document.body;
-  const offsetParentRect = node.offsetParent === document.body ? {left: 0, top: 0} : offsetParent.getBoundingClientRect();
+export function offsetXYFromParentOf(evt: {clientX: number, clientY: number}, node: HTMLElement & {offsetParent: HTMLElement}, draggableCore: DraggableCore): ControlPosition {
+  const offsetParent = draggableCore.props.offsetParent || node.offsetParent || document.body;
+  const offsetParentRect = offsetParent ? {left: 0, top: 0} : offsetParent.getBoundingClientRect();
 
   const x = evt.clientX + offsetParent.scrollLeft - offsetParentRect.left;
   const y = evt.clientY + offsetParent.scrollTop - offsetParentRect.top;
--- a/lib/utils/positionFns.es6
+++ b/lib/utils/positionFns.es6
@@ -66,7 +66,7 @@ export function canDragY(draggable: Draggable): boolean {
 export function getControlPosition(e: MouseEvent, touchIdentifier: ?number, draggableCore: DraggableCore): ?ControlPosition {
   const touchObj = typeof touchIdentifier === 'number' ? getTouch(e, touchIdentifier) : null;
   if (typeof touchIdentifier === 'number' && !touchObj) return null; // not the right touch
-  return offsetXYFromParentOf(touchObj || e, ReactDOM.findDOMNode(draggableCore));
+  return offsetXYFromParentOf(touchObj || e, ReactDOM.findDOMNode(draggableCore), draggableCore);
 }
 
 // Create an data object exposed by <DraggableCore>'s events


arekkas:
I don't believe the tests failing is my fault, but some message that changed:

```
	Expected spy error to have been called with [ 'Warning: Failed propType: Invalid prop transform passed to Draggable - do not set this, set it on the child.' ] but actual calls were [ 'Warning: Failed prop type: Invalid prop transform passed to Draggable - do not set this, set it on the child.
```

aeneasr:
The issue with the failing tests is that React has changed the way propType warnings are being thrown. It now provides a backtrace. On `master` the tests fail likewise. I am not sure how to fix this, because there are different error messages being thrown:

```
Expected spy error to have been called with

[ 'Warning: Failed prop type: Invalid prop transform passed to Draggable - do not set this, set it on the child.' ]

but actual calls were

[ 'Warning: Failed prop type: Invalid prop transform passed to Draggable - do not set this, set it on the child.

        in Draggable' ],

[ 'Warning: Failed prop type: Invalid prop transform passed to DraggableCore - do not set this, set it on the child.

        in DraggableCore (created by Draggable)

        in Draggable' ].
```

The only way I can think of right now is to replace `toHaveBeenCalledWith` with `toHaveBeenCalled`

acusti:
The tests are fixed by #176

aeneasr:
ah thanks, I will rebase once merged

STRML:
Thanks for this, I am reviewing. I'm looking for a simpler solution, though. Will get back to you shortly.

STRML:
I'd prefer an `offsetParent` prop rather than this boolean so it can be generic to the user's needs.

aeneasr:
@STRML sounds good, will do!

aeneasr:
Ok, it's rebased and `offsetParent chang eis included, thanks for the review :)

lib/DraggableCore.es6
```diff
@@ -67,6 +67,12 @@ export default class DraggableCore extends React.Component {
     enableUserSelectHack: PropTypes.bool,
 
     /**
+     * `offsetParent`, if set, uses the passed DOM node to compute drag offsets
+     * instead of using the parent node.
+     */
+    offsetParent: PropTypes.bool,
```
STRML: This isn't a bool. Unfortunately there's no dom node propType so it should be something like:

``` js
offsetParent: function(props, propName) {
  if (process.browser && props[propName] && props[propName].nodeType !== 1) {
    throw new Error("Draggable's offsetParent must be a DOM Node.");
  }
}
```

lib/DraggableCore.es6
```diff
@@ -152,6 +158,7 @@ export default class DraggableCore extends React.Component {
     cancel: null,
     disabled: false,
     enableUserSelectHack: true,
+    offsetParent: false,
```
STRML: This should be `null`

lib/DraggableCore.es6
aeneasr: Ah right, sorry I overlooked when changing that.

README.md
```diff
@@ -249,11 +249,12 @@ on itself and thus must have callbacks attached to be useful.
   cancel: string,
   disabled: boolean,
   enableUserSelectHack: boolean,
+  offsetParent: boolean,
```
STRML: As below, this should be an HTMLElement

STRML:
Please rebase again as #176 was merged.

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "react-draggable",
+  "name": "@brandcast/react-draggable",
```
arekkas: this doesn't look right

aeneasr:
Sorry, I updated the code quite hasty previously. Everything should be ok now, I hope. Thanks for your patience

lib/utils/domFns.es6
```diff
@@ -95,9 +96,9 @@ export function innerWidth(node: HTMLElement): number {
 }
 
 // Get from offsetParent
-export function offsetXYFromParentOf(evt: {clientX: number, clientY: number}, node: HTMLElement & {offsetParent: HTMLElement}): ControlPosition {
-  const offsetParent = node.offsetParent || document.body;
-  const offsetParentRect = node.offsetParent === document.body ? {left: 0, top: 0} : offsetParent.getBoundingClientRect();
+export function offsetXYFromParentOf(evt: {clientX: number, clientY: number}, node: HTMLElement & {offsetParent: HTMLElement}, draggableCore: DraggableCore): ControlPosition {
+  const offsetParent = draggableCore.props.offsetParent || node.offsetParent || document.body;
+  const offsetParentRect = offsetParent ? {left: 0, top: 0} : offsetParent.getBoundingClientRect();
```
STRML, line 101: This condition will always be true.

I'm going to refactor this a bit and push. Thanks for the contribution.

aeneasr, line 101: Perfect, thank you!

aeneasr, line 101: Ah right this should be `offsetParent === document.body`

STRML, line 101: Yep, but we can't even use document.body, we have to use `node.ownerDocument`.

Almost done, just writing a test to verify.

STRML:
Pushed in #179!

aeneasr:
Thanks @STRML !!

jpokorny-cz:
Hi everybody, 

can someone help me please how to use this feature? How can i pass html node, when it is not rendered yet?

```
  <GraphElement
            id={element.id}
            key={element.id}
            posX={element.cords.x}
            posY={element.cords.y}
            parrentOffset={ReactDom.findDOMNode(this)} // Will work after rerender ....
            onElementJointDragStart={() => this.props.onJointDragStart(element.id)}
            onElementJointDragMove={(event, data) => this.props.onJointDragMove(element.id, { x: data.x, y: data.y })}
          />
```

Thanks

acusti:
Like you pointed out, you need the parent to have been mounted before you can pass the node as `parentOffset`. Best method is to get the `ref` via the callback form, then either set the ref on state to trigger a rerender or use `this.forceUpdate`. Something like:

``` js
setRef = (ref) => {
  this.rootRef = ref;
  this.forceUpdate();
},

render() {
  // First comes your parent html; I just put a div as an example
  return (
    <div className="parent" ref={this.setRef}>
      { this.rootRef ? (
        <GraphElement
            id={element.id}
            key={element.id}
            posX={element.cords.x}
            posY={element.cords.y}
            parentOffset={this.rootRef}
            onElementJointDragStart={() => this.props.onJointDragStart(element.id)}
            onElementJointDragMove={(event, data) => this.props.onJointDragMove(element.id, { x: data.x, y: data.y })}
        />
      ) : null }
    </div>
  );
}
```

jpokorny-cz:
Thank you for quick replay.

It works. I just added a condition to prevent a loop.

```
setRef(ref) {
    if (this.graphDom === undefined) {
      this.graphDom = ref;
      this.forceUpdate();
    }
  }
```

But I am not sure if this is the correct way how to work with react.

STRML:
Usually you do something like this:

``` js
componentDidMount() {
  this.setState({mounted: true});
}

render() {
  ...
  ref="graphElement"
  parentOffset={this.state.mounted ? this.refs.graphElement : null}
  ...
}
```

acusti:
String refs are discouraged because of lack of composability / HOC flexibility ([reference](http://stackoverflow.com/a/39112162/333625))

rhalff:
Also wonder whether you actually need to pass the parentOffset.

The drag handler will also receive `deltaX` and `deltaY` and is probably what you need to calculate a new position.

``` js
onDrag (ev, data) {

  ....

  const coords = {
     x: old.x + data.deltaX,
     y: old.y + data.deltaY
  }
}
```

jpokorny-cz:
Thank you all for your tips and tricks. I think ref should be a function. Toggle the mounted variable in state seems to be a good solution. I need this cause I have draggable elements within another draggable element and all shoud have the same parentOffset.

PR #176, opened 2016-07-18
Fix propType tests for latest React (v15.2.0)

acusti:
React 15.2.0 changed the propType validation warning text:
https://github.com/facebook/react/blob/master/CHANGELOG.md#1520-july-1-2
016

This PR adapts the assertion to work for both versions of the warning text.


STRML:
Thanks!

PR #180, opened 2016-07-19
Use window/document relative to DOM element

STRML:
Use the ownerDocument and window of the DOM element being manipulated
rather than the global window and document objects. This makes
react-draggable work with a tool like
https://github.com/ryanseddon/react-frame-component

This replaces #177 which still needs some tests.

### DIFF
--- a/lib/utils/positionFns.es6
+++ b/lib/utils/positionFns.es6
@@ -17,15 +17,17 @@ export function getBoundPosition(draggable: Draggable, x: number, y: number): [n
   const node = ReactDOM.findDOMNode(draggable);
 
   if (typeof bounds === 'string') {
+    const {currentDocument} = node;
+    const currentWindow = node.defaultView;
     let boundNode;
     if (bounds === 'parent') {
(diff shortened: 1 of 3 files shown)


lib/utils/positionFns.es6
```diff
@@ -17,15 +17,17 @@ export function getBoundPosition(draggable: Draggable, x: number, y: number): [n
   const node = ReactDOM.findDOMNode(draggable);
 
   if (typeof bounds === 'string') {
+    const {currentDocument} = node;
```
acusti, line 20: `currentDocument` is a thing? Didn’t know about that one.

STRML, line 20: Thanks, it's actually not. Gap in our test coverage to be sure

STRML, line 20: https://github.com/mzabriskie/react-draggable/commit/ac269435565fe584f0409345dcb79e8f5b682828

acusti:
Beautiful! This is exciting.

PR #183, opened 2016-07-29
Fix for latest iPad scrolling instead of dragging vertically.

JaneCoder:
This is a fix for iPad, which has a flaw with vertical dragging related to the scroll bar. Other touch devices, even iPhone worked fine without the fix. This version uses the event helpers to add and remove the events.

### DIFF
--- a/lib/DraggableCore.es6
+++ b/lib/DraggableCore.es6
@@ -196,7 +196,6 @@ export default class DraggableCore extends React.Component {
   }
 
   handleDragStart: EventHandler<MouseEvent> = (e) => {
-
     // Make it possible to attach event handlers on top of this one.
     this.props.onMouseDown(e);
 
@@ -258,6 +257,7 @@ export default class DraggableCore extends React.Component {
   };
 
   handleDrag: EventHandler<MouseEvent> = (e) => {
+
     // Stop scrolling on touch devices while user is dragging as this is an issue for ipad.
     if (this.state.dragging && this.state.scrolling) {
       const {ownerDocument} = ReactDOM.findDOMNode(this);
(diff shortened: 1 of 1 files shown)


JaneCoder:
Unfortunately this version is failing for vertical drag on Lenovo touch screens in Firefox now. We will do some more testing and try to determine the cause.

STRML:
I appreciate your tenacity! Browsers implement this spec very badly.

JaneCoder:
This is looking good. We found that the tester had issues outside of our app with touch on chrome and firefox. Touch had to be re-enabled on her device. It was not related to the update to the ipad fix.

STRML:
I'm going to review this when I get home - have an iPad laying around - I'll update when I am sure this is working. Thanks!

JaneCoder:
Thank you for your wonderful work!

JaneCoder:
Hello @STRML. Have you had a chance to test with your ipad? Please let me know if there is anything different you'd like me to do.

nisjunk:
On iPadPro, all demos fail because the whole page also scrolls with the element. I think it is a show stopper issue.

JaneCoder:
nisjunk, we have that issue with the original version, but not with this patched version which stops scrolling on drag.

PS we are using a version of this pull request from our bitbucket repo, and will continue with our own patched version until this is accepted or someone else puts in a fix the owners like.

STRML:
Hi @JaneCoder,

Very sorry for how long it took to get to this.

I reviewed your fix and found that it fixed most use cases, but caused issues with bounded elements. After spending some time reviewing exactly how iOS handles touch events, I was able to very simply fix this in 8a51de9.

Thanks for reporting this!

STRML:
This is published and in `v2.2.3`.

JaneCoder:
Sweet! Thanks so very much!

Issue #186, opened 2016-08-01
Cannot read property 'getComputedStyle' of undefined


anibali:
Version: react-draggable 2.2.0
Browser: Google Chrome Linux 51.0.2704.106 (64-bit)
Error: `Cannot read property 'getComputedStyle' of undefined`
Source of error: https://github.com/mzabriskie/react-draggable/blob/ac269435565fe584f0409345dcb79e8f5b682828/lib/utils/positionFns.es6#L29

Looking a few lines up from the source of the error, I think that changing

``` js
const ownerWindow = node.defaultView;
```

to

``` js
const ownerWindow = ownerDocument.defaultView;
```

might fix the problem. I've never used `defaultView` before though, and I'm not sure what is trying to be achieved with it, so take that with a pinch of salt.

I have no problems with react-draggable 2.1.2.

grunnlock:
Happens to me as well when using the "bounds" property:
`<ReactDraggable
      bounds="body"
      handle=".dashboard-panel-header"
      onStart={ @focusPanel }
      onClick={ @focusPanel }
    >`

afilp:
It happens to me too, this is very problematic and I do not know what to do! I would appreciate a quick fix, thanks a lot!

CaptainChemist:
@anibali I ran into this same issue, did you end up fixing it? I too was able to fix it by reverting to an earlier version (2.1.2).

CaptainChemist:
The weird thing with my project is that even if I revert back to a previously working version on github and do a `rm -rf node_modules` followed by an npm install, it still gives me this error. I wonder if this problem is something other than the project itself- it also appears to be broken in safari as well as chrome for me.

anibali:
@CaptainChemist I ended up reverting to the earlier version. I did some poking around in the console and found that the window object was available via the document, but I didn't actually make the change. The line I'm referring to is https://github.com/mzabriskie/react-draggable/blob/ac269435565fe584f0409345dcb79e8f5b682828/lib/utils/positionFns.es6#L21

Essentially `node.defaultView` is `undefined`, and that's causing the problem.

vhmth:
Confirmed that this is the case for me as well. Reverting to 2.1.2 indeed fixed the issue.

jonshamir:
Any updates on this issue?

uriklar:
+1

STRML:
If you guys have an issue please submit a PR.

STRML:
[...]

PR #190, opened 2016-08-11
fix Cannot read property 'getComputedStyle' of undefined #186

nassimbenkirane:
Fix the problem that happened since version 2.2.0 when bounds property was used


STRML:
Fixed in 2.2.1 - would have accepted but built files were erroneously included in this commit.

nassimbenkirane:
Great ! Thanks for your reactivity 👍

Issue #162, opened 2016-05-25
Uncaught TypeError: Cannot read property 'SVGElement' of undefined


tlrobinson:
In v2.1.1 I'm getting this error when trying to use `Draggable`:

`react-draggable.js:267 Uncaught TypeError: Cannot read property 'SVGElement' of undefined`

It looks like the problem has to do with `global` being `undefined` due to `(function() { return this; })()` returning `null` rather than the global object in strict mode. I'm not really sure why it's being executed in strict mode though.

STRML:
That's odd. Strict mode is intended and implied in Babel modules, but Webpack should be handling global injection properly.

I don't see this issue on any of the demos. Perhaps it has something to do with your environment?

artemstorozhuk:
Exactly same issue for me. I've got no idea how to fix it.

STRML:
You could help by giving me reproduction steps.

artemstorozhuk:
Thanks for your quick response.

In console:
`npm i --save react-draggable`

In code:

```
import React from 'react'
import {render} from 'react-dom'
import Draggable from 'react-draggable';

render(
    <Draggable>
        <div>content</div>
    </Draggable>,
    document.getElementById('root')
);

```

Btw component is dragging properly.

amoilanen:
Might be that I found a workaround.

Observed the same error when the code of `react-draggable` was run through the Babel transform by Webpack while being located under 'node_modules'. After turning off the Babel transform the issue disappeared, now no errors in the console.

Looks like some Webpack + Babel issue.

PR #193, opened 2016-08-15
Avoid having 'global' 'undefined'

amoilanen:
Tested that this change fixes the issue https://github.com/mzabriskie/react-draggable/issues/162 "Uncaught TypeError: Cannot read property 'SVGElement' of undefined"

Original changes were made by tkrkt in https://github.com/tkrkt/react-draggable/commit/41ec001671dec5c1afebecb5a8e984a13010f6a4 in a fork of react-draggable

Alternative approach is to avoid using global and assume that SVGElement can be defined as a global, instead of:

``` javascript
    if(typeof global.SVGElement !== 'undefined' && ReactDOM.findDOMNode(this) instanceof global.SVGElement) {
    ...
```

we can use

``` javascript
    if(typeof SVGElement !== 'undefined' && ReactDOM.findDOMNode(this) instanceof SVGElement) {
    ...
```

Please, note that after running webpack in Windows at the root the dist file has been changed somewhat, looks like line-endings were changed and not sure if source map has been properly generated, might be that we may need to run webpack to generate the dist in a Linux or Mac environment.

### DIFF
--- a/dist/react-draggable.js
+++ b/dist/react-draggable.js
@@ -11,41 +11,41 @@
 return /******/ (function(modules) { // webpackBootstrap
 /******/ 	// The module cache
 /******/ 	var installedModules = {};
-/******/
+
 /******/ 	// The require function
 /******/ 	function __webpack_require__(moduleId) {
-/******/
+
 /******/ 		// Check if module is in cache
 /******/ 		if(installedModules[moduleId])
 /******/ 			return installedModules[moduleId].exports;
-/******/
+
 /******/ 		// Create a new module (and put it into the cache)
 /******/ 		var module = installedModules[moduleId] = {
 /******/ 			exports: {},
 /******/ 			id: moduleId,
 /******/ 			loaded: false
 /******/ 		};
-/******/
+
 /******/ 		// Execute the module function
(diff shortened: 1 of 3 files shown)


STRML:
A few things:
1. This error only manifests when using babel on your node-modules folder - don't do that, tons of libraries break when you do.
2. This feels like a hack to me. I don't think it makes sense to introduce a dependency (and look, it has its own dependencies!) just to try to access the window.
3. Please don't add built files to PRs.

amoilanen:
Hi, thank you for a quick reply.
1. It is true that this error only manifests itself when running node-modules through Babel however it might be necessary to do so if, for example, an older version of PhantomJS is used to run the unit tests.
2. Actually it is just explicitly requiring a used dependency instead of depending on the environment to provide it, i.e. instead of expecting to find 'SVGElement' in a window object we expect to find 'global' there.
3. There is a better solution: avoid referring to 'global'. If SVGElement is defined it should be defined in the global scope.

Updated the PR, removed the modifications to dist.

STRML:
Seems good. Thanks.

amoilanen:
Thanks

PR #198, opened 2016-09-07
Get ownerDocument before calling onStop handlers

acusti:
If a draggable `onStop` handler (invoked in [DraggableCore:318](https://github.com/mzabriskie/react-draggable/blob/master/lib/DraggableCore.es6#L318)) causes the wrapped component to unmount, calling `ReactDOM.findDOMNode` in [line 321](https://github.com/mzabriskie/react-draggable/blob/master/lib/DraggableCore.es6#L321) will error with `Uncaught Invariant Violation: findDOMNode was called on an unmounted component`.

Using findDOMNode before calling `onStop` avoids that bug and prevents having to also invoke it for the userSelect hack. A win-win.

### DIFF
--- a/lib/DraggableCore.es6
+++ b/lib/DraggableCore.es6
@@ -301,9 +301,10 @@ export default class DraggableCore extends React.Component {
     if (position == null) return;
     const {x, y} = position;
     const coreEvent = createCoreData(this, x, y);
+    const {ownerDocument} = ReactDOM.findDOMNode(this);
 
     // Remove user-select hack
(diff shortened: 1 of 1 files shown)


STRML:
:+1: Nice fix.

bkniffler:
Could you please publish this to npm? Thanks!

acusti:
@bkniffler If you want an installable version with this fix, you can temporarily use:
`"react-draggable": "Brandcast/react-draggable#1de1e66",`

in your package.json until this gets published. The `#1de1e66` hash just points to my [commit of the built files](https://github.com/Brandcast/react-draggable/commit/1de1e6609434ba3582c2115de3159597c7f82be1).

STRML:
This is published as v2.2.2.

PR #199, opened 2016-09-19
adds undocumented bounds option to docs

qamil:
adds docs for an crucial undocumented feature

https://github.com/mzabriskie/react-draggable/blob/1b9000e37bd683b6646afef23dad4a78f7017052/lib/utils/positionFns.es6#L26


STRML:
Thanks, this is helpful.

PR #224, opened 2017-02-09
add comma

sunopar:
add comma


STRML:
That's not how `import type` works - that is a Flow import.

PR #226, opened 2017-02-15
pass the bounded x and y in the DraggableData

edeustace:
I was running into an issue where I was trying to synch a component's position with another one that was being dragged via a `Draggable`, I was using `onDrag` for this. 

It works fine until you go outside the bounds at which point the x/y would fall out of synch. This PR returns the bounded x/y (set in `newState`) in `DraggableData` back to the callback.


romamatusevich:
@STRML, can these changes be added to the next release?

romamatusevich:
@STRML, could you please rerun tests? I think the failed checks may be an environment issue.

STRML:
Ported into 55f701e - thanks for the contribution.

Issue #119, opened 2015-12-23
Provide typescript definitions


Strate:
Hi!

In readme file you provide description of props, written in typescript notation. Could you please provide `d.ts` file?

STRML:
That's just habit from writing a lot of Flow code. Happy to accept a `d.ts` pull request, I don't use TypeScript personally.

erfangc:
Here is what I came up with, please give me your opinions, before we make this a PR to either this Repo or the [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) repo.

Something is definitely amiss, as the following only works if we ask TS to target ES6. Not sure what needs to be done to ensure more robust compatibility.

``` ts
///<reference path="typings/react/react.d.ts"/>

declare module 'react-draggable' {

    import {Component} from 'react';
    import Props = __React.Props;

    interface DraggableBounds {
        left: number
        right: number
        top: number
        bottom:number
    }

    interface DraggableProps extends Props<Draggable> {
        axis?: string
        bounds?: DraggableBounds|string|boolean
        start?:{x:number,y:number}
        zIndex?:number
    }

    interface DraggableCoreProps extends Props<DraggableCore> {
        allowAnyClick: boolean
        disabled: boolean
        enableUserSelectHack: boolean
        grid: number[]
        handle: string
        cancel: string
        onStart: (event, ui) => boolean
        onDrag: (event, ui) => boolean
        onStop: (event, ui) => boolean
        onMouseDown: (event, ui) => boolean
    }

    export default class Draggable extends Component<DraggableProps, any> {
    }

    export class DraggableCore extends Component<DraggableCoreProps, any> {
    }

}
```

``` json
{
    "compilerOptions": {
        "module": "es6",
        "target": "es6",
        "sourceMap": true,
        "jsx": "react"
    },
    "exclude": [
        "node_modules"
    ]
}
```

``` jsx
// sample code
///<reference path="react-draggable.d.ts"/>
import * as React from 'react';
import * as ReactDOM from 'react-dom';
import {default as Draggable, DraggableProps} from 'react-draggable';
import * as ReactDraggable from 'react-draggable';

class SampleComponent extends React.Component<any,any> {
    constructor(props:any) {
        super(props);
    }

    render() {

        const props:DraggableProps = {};
        debugger;
        return (
            <Draggable {...props}>
                <div>
                    <h1>Hello World!</h1>
                </div>
            </Draggable>
        )
    }
}
[...]

PR #229, opened 2017-03-06
Provide typescript definitions #119

erfangc:


erfangc:
not sure why the build is failing ... error is with Chromium ... seems like a issue that has occurred before by looking at the build history

didn't change code - just added definition file, so in theory no test can break, can we  trigger a simple rerun?

STRML:
:+1: Thanks!

jorgegonzalez:
I'm new to Typescript and I'm not exactly sure how to use this component. I've tried `import * as Draggable from 'react-draggable';` but that doesn't seem to work.

Issue #244, opened 2017-04-27
typescript errors in 2.2.4


timothyallan:
Getting a whack of TS errors now when using 2.2.4 and TypeScript 2.3
```
node_modules/react-draggable/index.d.ts
(49,1): error TS1046: A 'declare' modifier is required for a top level declaration in a .d.ts file.

node_modules/react-draggable/index.d.ts
(49,7): error TS2323: Cannot redeclare exported variable 'Draggable'.

/node_modules/react-draggable/index.d.ts
(49,66): error TS2304: Cannot find name 'DraggableState'.

node_modules/react-draggable/index.d.ts
(60,5): error TS2484: Export declaration conflicts with exported declaration of 'DraggableCore'.
```

lostfictions:
Bit by this as well! Just to be clear, it totally breaks compilation for anyone using the package with TypeScript (and probably throws some weird stuff for anyone using the package with vanilla JS or Babel in VS Code or other IDEs that consume TypeScript defs.) The only way to get it compiling again is to delete the .d.ts file from node_modules, which obviously doesn't work for redistribution.

I'm only just trying react-draggable with TypeScript for the first time today, but looking at the definitions I'm not sure how they ever worked since they [reference an identifier that doesn't exist](https://github.com/mzabriskie/react-draggable/blob/master/index.d.ts#L49-L55).

Don't have time to submit a PR today, but [here's a gist of a fixed .d.ts file](https://gist.github.com/lostfictions/f0b4e6ae62d6f102a4d04b6d926da9ab).

I could also submit this as a PR to DefinitelyTyped instead and react-draggable could cut a new release with the definitions removed from this repo? I really appreciate when maintainers are willing to host definitions in the repo since it reduces friction for consumers of the package (and automagically improves completions in editors like VS Code even for vanilla JS!) That said, I think it might only makes sense to host definitions in-repo if maintainers are up for keeping them up-to-date and testing against TypeScript when the API changes.

STRML:
Thanks @lostfictions. I've published v2.2.5 which fixes this issue.

I use Flow in our own projects, not TS. Perhaps your DT idea is better. But if I wanted to ensure that the TS definitions are correct in the future and maintain this definition, do you have suggestions on how to add that to the test suite?

lostfictions:
Thanks a lot for the quick turnaround!

[...]

PR #245, opened 2017-04-29
Move TypeScript defs to own folder and add tests

lostfictions:
Here's a PR for testing TS defs, as discussed in #244. It doesn't test them against the existing spec file -- between Karma augmenting the global namespace and the spec importing some Flow files, it ended up being a bit too tricky. Instead, this PR adds a `tests.tsx` file that simply verifies that the correct props are supported (as well as being optional.)

(When Microsoft/TypeScript#15160 lands it'll be possible to enforce that `Draggable` and `DraggableCore` have a child as well.)

I've also added invoking `tsc` to perform the checking as part of the `test` task in `Makefile`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,7 @@ dist/%.js: $(BIN)
 
 test: $(BIN)
 	@$(BIN)/karma start --single-run
+	@$(BIN)/tsc -p typings
 
 dev: $(BIN)
 	script/build-watch
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "build": "make clean build",
     "lint": "make lint"
   },
-  "typings": "./index.d.ts",
+  "typings": "./typings/index.d.ts",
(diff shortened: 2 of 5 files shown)


lostfictions:
(Incidentally, spotted some invalid/deprecated props in the spec while I was looking things over:

- `zIndex` here is deprecated: https://github.com/mzabriskie/react-draggable/blob/master/specs/draggable.spec.jsx#L118
- `style` here should be on the child: https://github.com/mzabriskie/react-draggable/blob/master/specs/draggable.spec.jsx#L629

Not sure why these aren't throwing via PropTypes verification but I don't have any more time to investigate this, sorry :(  Just thought I'd point them out!)

STRML:
Thanks. I removed the root `index.d.ts` and updated the lockfile. I tested a few invalid definitions and it threw as expected. I appreciate it a lot!

PR #254, opened 2017-05-26
make Draggable explicit default export

HBehrens:
When I was trying to use this package with my TypeScript setup I ran into the issue that

    import Draggable from 'react-draggable';

resulted in `Draggable === undefined` as TypeScript expects default exports to be assigned as `module.exports.default = ...`.  While I could probably change `typings/index.d.ts` to match the actual behavior `import * as Draggable from 'react-draggable'` I understand that the import statement above is the desired behavior.

This PR changes `index.js` to meet TS' expectation but doesn't remove the current `module.exports = ...` to avoid breaking compatibility.

I am not sure how exactly I could provide a unit-test for this. Any help appreciated.


r0stig:
Just commenting to show my interest in this PR getting merged.

HBehrens:
> Just commenting to show my interest in this PR getting merged.

How can I help with it?

STRML:
Fixed in afd7721

PR #261, opened 2017-06-29
Scroll works in Google Chrome. We have to use transparent selection…

dherbolt:
… instead of user-select:none

### DIFF
--- a/lib/utils/domFns.es6
+++ b/lib/utils/domFns.es6
@@ -1,6 +1,6 @@
 // @flow
 import {findInArray, isFunction, int} from './shims';
-import browserPrefix, {getPrefix, browserPrefixToStyle, browserPrefixToKey} from './getPrefix';
+import browserPrefix, {browserPrefixToKey} from './getPrefix';
 
 import type {ControlPosition} from './types';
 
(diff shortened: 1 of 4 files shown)


STRML:
Is there a reason to restrict this to just Chrome? It appears `::selection` has [good support](https://developer.mozilla.org/en-US/docs/Web/CSS/::selection).

dherbolt:
@STRML I think no. And of course it will keep the code base simple... So I will  change it for all browsers, ok?

STRML:
Yep, that sounds good, and a lot less hacky than my current implementation that slices the `document.body.style` string.

dherbolt:
Ok, give me some time to do it.

dherbolt:
Hi, the Travis CI build failed but I can't find the reason. Can anyone help me, please?

STRML:
I've seen spurious CI errors on this project, has something to do with the page spontaneously reloading, but only occasionally. I restarted the build and it worked fine.

lib/utils/domFns.es6
```diff
@@ -127,21 +127,24 @@ export function getTouchIdentifier(e: MouseTouchEvent): ?number {
...
-  const style = body.getAttribute('style') || '';
-  if (userSelectReplaceRegExp.test(style)) return; // don't add twice
-  body.setAttribute('style', style + userSelectStyle);
+export function addUserSelectStyles(doc: Document) {
+  let styleEl = doc.getElementById('react-draggable-style-el');
+  if (!styleEl) {
+    styleEl = doc.createElement('style');
+    styleEl.type = 'text/css';
+    styleEl.id = 'react-draggable-style-el';
+    styleEl.innerHTML = '.react-draggable-transparent-selection *::-moz-selection {background: transparent;}\n';
```
STRML, line 138: Will other browsers require prefixes or can we get away with just this?

lib/utils/domFns.es6
```diff
@@ -127,21 +127,24 @@ export function getTouchIdentifier(e: MouseTouchEvent): ?number {
...
+    styleEl.innerHTML += '.react-draggable-transparent-selection *::selection {background: transparent;}\n';
+    doc.getElementsByTagName('head')[0].appendChild(styleEl);
+  }
+  doc.body.classList.add('react-draggable-transparent-selection');
```
STRML: Unfortunately we can't count on this existing (although it does in every modern browser IE11+). My preference would be to munge the string so we maintain IE9 support.

specs/draggable.spec.jsx
```diff
@@ -89,7 +73,7 @@ describe('react-draggable', function () {
           <div
             className="react-draggable"
             style={{
-              [transformKey]: 'translate(0px, 0px)'
+              'transform': 'translate(0px, 0px)'
```
STRML: Technically this should remain - and I need to fix other parts of the test - reason being that we could (and should) add an old browser like IE9 to the suite.

STRML (changes requested):
Some minor nits re: IE9 - otherwise thanks for the contribution!

lib/utils/domFns.es6
dherbolt: We need vendor prefix only for Firefox, see [::selection](http://caniuse.com/#search=%3A%3Aselection)

dherbolt:
@STRML The Travis CI failed again. Can you restart it, please?

STRML:
Merged into master w/ f33352ec2aabebf02c3073f9ccd4a78f335483b0, thanks!

lib/utils/domFns.es6
```diff
@@ -127,21 +127,24 @@ export function getTouchIdentifier(e: MouseTouchEvent): ?number {
...
+  }
+  addClassName(doc.body, 'react-draggable-transparent-selection'); 
 }
 
-export function removeUserSelectStyles(body: HTMLElement) {
-  const style = body.getAttribute('style') || '';
-  body.setAttribute('style', style.replace(userSelectReplaceRegExp, ''));
+export function removeUserSelectStyles(doc: Document) {
+  removeClassName(doc.body, 'react-draggable-transparent-selection');
+  window.getSelection().removeAllRanges();  // remove selection caused by scroll
```
grg-akshay, line 147: @STRML @dherbolt Can you please tell me how can scrolling cause selection? And why this removal is important?

dherbolt: When you hold mouse button down to drag some element it cases also invisible (transparent) selection of elements on the page. When you reach bottom/top border of the browser window, this selection starts scrolling with the content in the window (as when you are selecting text).
When you leave mouse button, this selection used to scroll with the page should be removed otherwise it will be probably visible in some cases or can cause somethig else. It isn't nothing else then final cleanup.

VincentCharpentier: Hey @dherbolt,
Sorry to ping on something old but, does that mean the auto-scroll feature relies on the hidden selection ?

dherbolt: Yes, exactly.

Issue #223, opened 2017-02-06
Completely center element conflicting with applied transform


scniro:
I had a modal leveraging this library for dragability. The size of this modal is unknown, so I initialize the style with the following rule...

    transform: translate(-50%, -50%);

this will dead center my element no matter the size. However, as soon as I start to drag, the transform kicked in by this library starts back at `translate(0, 0);` - causing the element to jump at first, then drag accordingly. 

I looked into the options such as `defaultPosition` and this works well if the height/width are *known*. Ideally, I'd like to pass percents to it, but that results in an error.

Is there a workaround for this? I've already searched through all the issues and haven't seen one that has this exact problem. Or, how can I *exactly* center an element, and begin the dragging according with the 50% offset taken into account to where the element starts on a grid?

STRML:
Wrap your element in another element, so that the centering hack you're using does not conflict with `<Draggable>` or use another of the many options out there for centering the element (such as flexbox or margin: auto).

eric-burel:
Hi, translating the wrapped component won't work, as the Draggable zone itself won't be translated. We need everything to be translated in this case.

Sadly, there seems to be no options to center a floating element whose width is unknown on the sreen than translating it from -50% after positionning it.
The use case is limited but legitime.

Therefore the solution is indeed to accept a string as `defaultPosititon` param. Then, we can use `calc` to compute the translation, something like this: `transform(calc(defaultPosition.x + x), calc(defaultPosition.y +y))`

PR #271, opened 2017-08-17
[POC] Allow string default position

eric-burel:
Hi,
Since `react-draggable` relies on transform, it can sometimes be difficult to create components that rely themselves on transform, since transformations will be overidden by react-draggable.

This PR especially solves #223. Indeed, the only way to center a floating div on the screen is to use a `transform(-50%, -50%)`.

So, I added the possibility to provide the `defaultPosition` as a string.

Cool stuffs:
- we can provide a relative `defaultPosition`, this is good for centering
- it does the job

Bad stuffs:
- it breaks automatic tests and sorry, I lack the time to update them (the live example is working though)
- it relies on `calc`, I am not sure if this is good / cross-browser / IE compatible
- it add computations in the `transform` generation, which reduce performance (but is it significant ?), instead of using the `defaultPosition` only during the component construction.
- it is not optimized, I put calc everytime even if all values are string


STRML:
Hi - apologies that `master` is now conflicting, could you make the following changes:

1. Don't commit `dist/`, and
2. Port your changes from the old `.es6` files (this extension had issues with some tools like `flow-upgrade`) over to `.js`.

Thanks for your contribution! :+1:

tnrich:
@eric-burel can you please finish this PR? I would also like to see this work! 

@STRML please please make this PR work

tnrich:
I made https://github.com/mzabriskie/react-draggable/pull/361 which fixes @STRML's above comments.

PR #281, opened 2017-09-11
Add peerDeps to package.json

amilajack:
Peer deps are exact same semver as devDeps

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,10 @@
     "webpack": "^3.5.5",
     "webpack-dev-server": "^2.7.1"
   },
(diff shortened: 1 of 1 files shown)


STRML:
Would actually just be `15.x` - we work with older minor versions w/o issue.

STRML:
:) It's just `15.x`, AFAIK `^15.x` is invalid.

STRML:
`peerDependencies` are generally a pain in the butt (they rarely tell you anything important, but print a lot of log noise) - going to close this and avoid adding them.

PR #217, opened 2017-01-05
allowing the ability to pass a scale parameter so it drags correctly …

Madisonw:
…at any zoom level

PR #284, opened 2017-09-13
Add ability to pass in scale prop to correctly move draggable element at different scales

commanderking:
Based on this original PR: https://github.com/mzabriskie/react-draggable/pull/217/files, but removed conflicts with dist files. 

Also, added an additional test to check for scale prop and an example for the example page. Wasn't sure if we wanted the scale in DraggableCore, but can add it if desired. 

@STRML, let me know if any changes are needed and I can get them in Thanks for maintaining this awesome library!

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,13 +217,15 @@ onStop: DraggableEventHandler,
 // becomes 'controlled' and is not responsive to user input. Use `position`
 // if you need to have direct control of the element.
 position: {x: number, y: number},
-}
 
 // Specifies the scale of the canvas you are dragging this element on. This allows
 // you to, for example, get the correct drag deltas while you are zoomed in or out via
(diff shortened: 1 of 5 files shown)


ibio:
This is really good idea! I'd like to use this feature as well.

commanderking:
Hi @STRML, wondering whether there are any changes that need to be made to this PR for it to get merged in. Thanks again for all your work in maintaining this library!

maixing:
When to add scale attribute?

arnauddb:
Can we merge this PR ?

STRML:
I'm wondering if this can be done automatically by querying the element's own effective scale. This could be on an on/off switch but would prevent the need for manually tweaking scale.

PR #285, opened 2017-09-15
fixed sticky click on IE11/Edge

ibio:
Using IE, Edge(13, 14 and 15), if this react-draggable is in an iframe, the draggable item gets 'stuck' on the mouse if the user attempts to drag it off iframe/screen.

To reproduce:
1 Make sure react-draggable is in an iframe
2 Open IE browser 
3 Try to click/drag the draggable item out of the iframe
4 Release mouse click(outside of the iframe). Observe as the user now moves mouse around the screen, the draggable item follows as if it's stuck to the pointer.


ibio:
@STRML hey, can you please have a look this PR? Thanks :)

STRML:
I don't believe this is guaranteed to succeed if `top` is cross-origin, which will break this for other users.

ibio:
Good point! if I use `try catch` first, and use `ownerDocument` as a backup, will that be acceptable?

ibio:
Use a new PR to track this issue.

Issue #283, opened 2017-09-12
Errors with Flow 0.54


EvHaus:
When using react-draggable and Flow 0.54 I get the following Flow type errors:

```
Error: node_modules/react-draggable/lib/Draggable.js:257
257:       [newState.x, newState.y] = getBoundPosition(this, newState.x, newState.y);
            ^^^^^^^^^^ unsupported expression pattern in destructuring

Error: node_modules/react-draggable/lib/Draggable.js:257
257:       [newState.x, newState.y] = getBoundPosition(this, newState.x, newState.y);
                        ^^^^^^^^^^ unsupported expression pattern in destructuring

Error: node_modules/react-draggable/lib/DraggableCore.js:322
322:         this.handleDragStop(new MouseEvent('mouseup'));
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^ MouseEvent. This type is incompatible with the expected param type of
 29: export type MouseTouchEvent = MouseEvent & TouchEvent2;
                                                ^^^^^^^^^^^ TouchEvent2. See: node_modules/react-draggable/lib/utils/types.js:29

```

akkie:
With flow 0.45.1 I get also some more errors:

```
Error: node_modules/react-draggable/lib/Draggable.js:257
257:       [newState.x, newState.y] = getBoundPosition(this, newState.x, newState.y);
            ^^^^^^^^^^ unsupported expression pattern in destructuring

Error: node_modules/react-draggable/lib/Draggable.js:257
257:       [newState.x, newState.y] = getBoundPosition(this, newState.x, newState.y);
                        ^^^^^^^^^^ unsupported expression pattern in destructuring

Error: node_modules/react-draggable/lib/DraggableCore.js:322
322:         this.handleDragStop(new MouseEvent('mouseup'));
                                 ^^^^^^^^^^^^^^^^^^^^^^^^^ MouseEvent. This type is incompatible with the expected param type of
 29: export type MouseTouchEvent = MouseEvent & TouchEvent2;
                                                ^^^^^^^^^^^ TouchEvent2. See: node_modules/react-draggable/lib/utils/types.js:29

Error: node_modules/react-draggable/lib/DraggableCore.js:328
328:         event.initMouseEvent('mouseup', true, true, window, 0, 0,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

[...]

PR #286, opened 2017-09-18
Fix errors with flow v0.54

hemnstill:
Suggested fix for issue #283

### DIFF
--- a/lib/DraggableCore.js
+++ b/lib/DraggableCore.js
@@ -106,7 +106,7 @@ export default class DraggableCore extends React.Component<DraggableCoreProps, D
      * instead of using the parent node.
      */
     offsetParent: function(props, propName) {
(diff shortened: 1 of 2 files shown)


lib/DraggableCore.js
```diff
@@ -106,7 +106,7 @@ export default class DraggableCore extends React.Component<DraggableCoreProps, D
      * instead of using the parent node.
      */
     offsetParent: function(props, propName) {
-      if (process.browser && props[propName] && props[propName].nodeType !== 1) {
+      if (process.browser != null && props[propName] && props[propName].nodeType !== 1) {
```
STRML: This should be `=== true`

hemnstill:
it has become `=== true`.

STRML:
Thanks!

PR #288, opened 2017-09-18
Fix Iframe sticky click with IE11/Edge

ibio:
When using IE, Edge(13, 14 and 15), if this react-draggable is in an iframe, the draggable item gets 'stuck' on the mouse if the user attempts to drag it off iframe/screen.

To reproduce:
1 Make sure react-draggable is in an iframe
2 Open IE browser
3 Try to click/drag the draggable item out of the iframe
4 Release mouse click(outside of the iframe). Observe as the user now moves mouse around the screen, the draggable item follows as if it's stuck to the pointer.

### DIFF
--- a/lib/DraggableCore.js
+++ b/lib/DraggableCore.js
@@ -289,8 +289,13 @@ export default class DraggableCore extends React.Component<DraggableCoreProps, D
     // this element. We use different events depending on whether or not we have detected that this
     // is a touch-capable device.
     addEvent(ownerDocument, dragEventFor.move, this.handleDrag);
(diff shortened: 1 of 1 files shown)


ibio:
@STRML I've added try catch to deal with `cross-origin` situation. Please have a look the latest change and let me know what you think. Thanks :)

STRML:
Closing this PR as stale. This PR targeted IE11/Edge (pre-Chromium) which are no longer supported browsers. The issue is no longer relevant. Thank you for the contribution!

PR #293, opened 2017-10-18
Add live demo

ipiranhaa:
Add some live demo to README.md


STRML:
Gif appears to be dead.

PR #303, opened 2017-11-27
Fix 'Cannot read property 'call' of undefined' in matchesSelector

svsool:
Also it would be better to type `el: Node` as `el: Element` https://github.com/svsool/react-draggable/blob/e1f7f45cb9e86212e80bd85da8cbef9697783886/lib/utils/domFns.js#L8, because Node interface doesn't provide any matches function see https://developer.mozilla.org/en-US/docs/Web/API/Node comparing to https://developer.mozilla.org/en-US/docs/Web/API/Element/matches


STRML:
Fixed in 176f691

PR #327, opened 2018-03-30
Build umd and min bundles with rollup

TrySound:
before

react-draggable.js 90.26K
react-draggable.min.js 45.36K

after

react-draggable.js 73.52K
react-draggable.min.js 15.45K

/сс @STRML


STRML:
Wow, that's quite an improvement.

FWIW, tested with Webpack 4 which got it from 45K minified build to 22K.

STRML:
Any reason to not add another `sizeSnapshot()` on the prod build?

TrySound:
No reason. Added.

STRML:
Sweet. This is pretty nice, I can't get down to 15KB no matter what I try with Webpack. But Webpack 4 is definitely better than 3 (45KB vs 22KB).

TrySound:
Would be good to release this as patch or minor and then I want to propose breaking change with named only exports to prevent [this hacks](https://github.com/mzabriskie/react-draggable/blob/master/index.js).

PR #328, opened 2018-04-02
Prefer named exports over default

TrySound:
This will be a breaking change, but I think it makes things cleaner. Also this allows as to distribute esm and cjs instead of umd as entry points.


TrySound:
Friendly ping

/cc @STRML

STRML:
This could break a lot of libs. I would strongly prefer to keep even something that is ugly rather than break backcompat.

TrySound:
Can't we just make major release?

TrySound:
I mean something ugly in esm will also break a lot of modules.

TrySound:
I saw in a lot of code 
```js
import { DraggableCore } from 'react-draggable';
```
and it will be broken with esm format in react-draggable

STRML:
Why would it break? I'm unfamiliar with releasing for esm given that it is not yet in wide use in node.

TrySound:
@STRML esm is wide use in bundling. The problem is that with this vendor

```js
const Draggable = 1
Draggable.DraggableCore = 2

export default Draggable
```
is not possible to do this because `DraggableCore` is not named export
```js
import { DraggableCore } from 'react-draggable';
```

If we produce it as named export
```js
const Draggable = 1
const DraggableCore = 2
Draggable.DraggableCore = DraggableCore

export default Draggable
export {
  DraggableCore
}
```

then cjs output will look like
```js
exports.default = 1
exports.default.DraggableCore = 2
exports.DraggableCore = 2
```

so cjs will be broken.


The only way to fix this collision is to get rid from default exports

TrySound:
Could you merge this?

STRML:
I'm confused. Are we not waiting on:

> Just wait please until I provide cjs and esm bundles in the next PR.

TrySound:
:)
> in the next PR

TrySound:
I meant wait to release

PR #329, opened 2018-04-02
Remove process.browser which is missing in browser

TrySound:
Upgraded flow and found this bug. It should fail in browser. Previously webpack added process polyfill which also increased bundle size.


TrySound:
Friendly ping

/cc @STRML

PR #331, opened 2018-04-11
Fixes some issues in the type definitions for TypeScript

matoilic:
The current type definitions use native events as parameter type for event listeners. But what is actually passed are instances of React's synthetic events. Also, depending on the input mode either a MouseEvent or a TouchEvent is passed.

The `package.json` currently only has the `typings` property which was used by older versions of TypeScript and leads to issues with current TypeScript versions where it randomly doesn't find the type definitions. Adding a `types` property to indicate the entry file for the type definitions fixes this. This is also what's recommended in the official docs when it comes to publishing NPM packages https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
     "flow": "flow"
   },
   "typings": "./typings/index.d.ts",
+  "types": "./typings/index.d.ts",
   "repository": {
     "type": "git",
     "url": "https://github.com/mzabriskie/react-draggable.git"
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -18,7 +18,7 @@ declare module 'react-draggable' {
     position: ControlPosition
   }
 
-  export type DraggableEventHandler = (e: MouseEvent, data: DraggableData) => void | false;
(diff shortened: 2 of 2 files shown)


matoilic:
The tests seem not to be set up properly to run on Windows. That's why everything fails on AppVeyor.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-draggable",
-  "version": "3.0.5",
+  "version": "3.0.5-pl.0",
```
STRML: Please do not change the version string.

STRML:
From the TS docs:

> Note that the "typings" field is synonymous with "types", and could be used as well.

Is this not the case?

package.json
matoilic: Sorry, I committed that by mistake. It was only for testing purposes.

matoilic:
I read that in the docs too. But before adding the `types` property the compilation would randomly fail and complain that it couldn't find `react-draggable`. After adding it to `package.json` the build was stable.

STRML:
Thanks for the PR.

PR #338, opened 2018-04-30
Perfer named exports and bundle esm and cjs

TrySound:
In this diff I replaced default export with named exports for better
tooling support.

Now both `Draggable` and `DraggableCode` have the only one way to import
it.

```js
import { Draggable, DraggableCore } from 'react-draggable';
```

or with commonjs

```js
const { Draggable, DraggableCore } = require('react-draggable');
```

Added esm and cjs bundles without bundled vendor dependencies. This
should have 8KB reduction for consumers.

Removed `browser` field in package.json because it's usually useless and
prefered over `module` which is not what we are trying to achieve.


TrySound:
Friendly ping. This PR is ready.

Issue #279, opened 2017-09-07
Safari for iDevices drags entire screen


scbrady:
When dragging in Safari on iDevices (I specifically tested an iPhone 7 and iPad Air 2, both running iOS 10.3) the entire screen scrolls with the drag. I tested it both on my website and on the react-draggable example page.

I see that in issue #165, this should have been taken care of, but it doesn't seem to be. Possibly Safari updated the way they handle dragging?

I'm going to keep looking for workarounds for this, but so far I haven't been able to find anything that works.

denis-sokolov:
`onStart={e => e.preventDefault()}` fixed it for me on an iPad Air iOS 10.3.

demian85:
This is not working for me on iOS 11 😕 
Any suggestions?

jasoniangreen:
I fixed by putting this
```
handlePreventTouchmoveWhenPanning = (event: *) => {
 if (this.state.panning) {
   event.preventDefault();
 }
};

componentDidMount () {
 window.document.body.addEventListener(‘touchmove’, this.handlePreventTouchmoveWhenPanning, {
   passive: false
 });
}

componentWillUnmount () {
 window.document.body.removeEventListener(‘touchmove’, this.handlePreventTouchmoveWhenPanning, {
   passive: false
 });
}
```

Issue #227, opened 2017-02-24
touch-action: none prevents scrolling in mobile Chrome when not on drag handle


DomBlack:
If you are using a handle I would only expect react-draggable to prevent touch events on the handle, instead it applies `touch-action: none` to the container, meaning anything inside it (handle or not) cannot receive touch events.

This makes browser scrolling impossible, could you give us a way to disable this behaviour and allow us to manually enable it directly on the drag handle?

slnsrn:
did you find a solution? :)

ZechyW:
As a workaround, you can override `touch-action` on the main container and drag handle by assigning classes like the following:
```css
/* Assign to the main container */
.drag-container{
  touch-action: auto !important;
}

/* Set as the handle class */
.drag-handle {
  touch-action: none;
}
```
This should allow scrolling/pinch-to-zoom/etc. on the container while also allowing the handle to initiate dragging (also see #51).

PR #351, opened 2018-07-20
Fixes touch scrolling for Safari and event listener error on Chrome

rskarras:
Issue #227
This addresses a breaking change with newer versions of Chrome (Chrome 56 and newer) which ignores preventDefault in passive event listeners. This resulted in the following error:  

> Unable to preventDefault inside passive event listener invocation.

There is documentation on this change at
[Making touch scrolling fast by default](https://developers.google.com/web/updates/2017/01/scrolling-intervention)

This addresses issue #227 by disabling scrolling at the beginning of a drag action rather than based on styling of an element.

Issue #279
Disables touch scrolling during drag events on newer versions of Safari.

### DIFF
--- a/lib/DraggableCore.js
+++ b/lib/DraggableCore.js
@@ -231,7 +231,7 @@ export default class DraggableCore extends React.Component<DraggableCoreProps, D
   handleDragStart: EventHandler<MouseTouchEvent> = (e) => {
 
     // Disable touch scrolling while dragging on mobile devices.
-    document.ontouchmove = function(e){ e.preventDefault(); };
+    e.target.ontouchmove = function(){ return false; };
 
     // Make it possible to attach event handlers on top of this one.
     this.props.onMouseDown(e);
@@ -340,10 +340,7 @@ export default class DraggableCore extends React.Component<DraggableCoreProps, D
 
   handleDragStop: EventHandler<MouseTouchEvent> = (e) => {
     if (!this.state.dragging) return;
-
-    // Reactivate touch scrolling after dragging action is complete on mobile devices.
-    document.ontouchmove = function(){ return true; };
-
+    
     const position = getControlPosition(e, this.state.touchIdentifier, this);
     if (position == null) return;
     const {x, y} = position;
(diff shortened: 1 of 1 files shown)


lib/DraggableCore.js
```diff
@@ -229,6 +229,10 @@ export default class DraggableCore extends React.Component<DraggableCoreProps, D
   }
 
   handleDragStart: EventHandler<MouseTouchEvent> = (e) => {
+
+    // Disable touch scrolling while dragging on mobile devices.
+    document.ontouchmove = function(e){ e.preventDefault(); };
```
STRML: This makes an assumption that the developer isn't using `document.ontouchmove` for something else. Can this be done via addEventListener() instead?

STRML:
```js
const PREVENT_DEFAULT = (e) => e.preventDefault();
// ...
window.addEventListener('touchstart', PREVENT_DEFAULT, { passive: false } );
// ...
window.removeEventListener('touchstart', PREVENT_DEFAULT);
```

rskarras:
Hi @STRML thanks for your comments and for reviewing this so quickly! You bring up a good point about document.ontouchmove being used elsewhere.

I have changed document.ontouchmove to e.target.ontouchmove and I believe that should solve this issue. 

{ return false; } is responsible for both of the following:
preventDefault - to disable scrolling following handleDragStart.
stopPropagation - to prevent the event from bubbling back up to the document level.

While I am sure it can be done with addEventListener() as suggested, it may not be necessary. 
Looking forward to hearing your thoughts on this. Thanks for your insight!

Cheers,

Sam

lib/DraggableCore.js
```diff
@@ -229,6 +229,10 @@ export default class DraggableCore extends React.Component<DraggableCoreProps, D
   }
 
   handleDragStart: EventHandler<MouseTouchEvent> = (e) => {
+
+    // Disable touch scrolling while dragging on mobile devices.
+    e.target.ontouchmove = function(){ return false; };
```
STRML, line 234: This still is not removed afterwards, so it's not workable as written.

STRML:
Fixed in #470

PR #359, opened 2018-08-20
Fix compare where portal elements are different instance to main window

jasonmarin42:
Fixes an issue where if you use the react portals with a window.open, the HTMLElement in the window is a different instance to the one in the main window, meaning you would always get an error.


STRML:
Great catch, thanks.

PR #361, opened 2018-09-01
defaultPosition now allows strings

tnrich:
This is a copy of https://github.com/mzabriskie/react-draggable/pull/271

but with the changes @STRML requested

### DIFF
--- a/lib/utils/domFns.js
+++ b/lib/utils/domFns.js
@@ -2,7 +2,7 @@
 import {findInArray, isFunction, int} from './shims';
 import browserPrefix, {browserPrefixToKey} from './getPrefix';
 
-import type {ControlPosition, MouseTouchEvent} from './types';
+import type {ControlPosition, DefaultControlPosition, MouseTouchEvent} from './types';
 
 let matchesSelectorFunc = '';
 export function matchesSelector(el: Node, selector: string): boolean {
@@ -109,22 +109,27 @@ export function offsetXYFromParent(evt: {clientX: number, clientY: number}, offs
   return {x, y};
 }
 
-export function createCSSTransform({x, y}: {x: number, y: number}, defaultPosition: {x: number | string, y: number | string }): Object {
-  const defaultX = (typeof defaultPosition.x === 'string') ? defaultPosition.x : defaultPosition.x + 'px';
-  const defaultY = (typeof defaultPosition.y === 'string') ? defaultPosition.y : defaultPosition.y + 'px';
-  const translation = (defaultPosition && defaultPosition.x !== 0 && defaultPosition.y !== 0)
-  ? `translate(${defaultX}, ${defaultY}) translate(${x}px,${y}px)`
-  : 'translate(' + x + 'px,' + y + 'px)';
-  // Replace unitless items with px
+export function createCSSTransform({x, y}: {x: number, y: number}, defaultPosition: DefaultControlPosition): Object {
+  let translation;
+  if (defaultPosition && (defaultPosition.x !== 0 || defaultPosition.y !== 0)) {
+    const defaultX = `${(typeof defaultPosition.x === 'string') ? defaultPosition.x : defaultPosition.x + 'px'}`;
+    const defaultY = `${(typeof defaultPosition.y === 'string') ? defaultPosition.y : defaultPosition.y + 'px'}`;
+    translation = `translate(${defaultX}, ${defaultY}) translate(${x}px,${y}px)`;
+  } else {
+    translation = `translate(${x}px,${y}px)`;
+  }
   return {[browserPrefixToKey('transform', browserPrefix)]: translation };
 }
 
-export function createSVGTransform({x, y}: {x: number, y: number}, defaultPosition: {x: number | string, y: number | string}): string {
-  const defaultX = (typeof defaultPosition.x === 'string') ? defaultPosition.x : defaultPosition.x + 'px';
-  const defaultY = (typeof defaultPosition.y === 'string') ? defaultPosition.y : defaultPosition.y + 'px';
-  const translation = (defaultPosition && defaultPosition.x !== 0 && defaultPosition.y !== 0)
-  ? `translate(${defaultX}, ${defaultY}) translate(${x}px,${y}px)`
-  : 'translate(' + x + ',' + y + ')';
+export function createSVGTransform({x, y}: {x: number, y: number}, defaultPosition: DefaultControlPosition): string {
+  let translation;
+  if (defaultPosition && (defaultPosition.x !== 0 || defaultPosition.y !== 0)) {
+    const defaultX = (typeof defaultPosition.x === 'string') ? defaultPosition.x : defaultPosition.x;
+    const defaultY = (typeof defaultPosition.y === 'string') ? defaultPosition.y : defaultPosition.y;
+    translation = `translate(${defaultX}, ${defaultY}) translate(${x},${y})`;
+  } else {
+    translation = `translate(${x},${y})`;
+  }
   return translation;
 }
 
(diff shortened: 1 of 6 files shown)


tnrich:
@STRML @eric-burel I fixed a couple issues in the commit I just made. Linting and tests now pass and I added a test for the new use case.

tnrich:
@bokuweb when this PR gets merged this issue https://github.com/bokuweb/react-rnd/issues/437 will no longer be relevant. I should just be able to pass "50%" for both x and y defaultPositions

tnrich:
@STRML any idea why this would be failing only on IE? 

That's the new spec I added: 
IE 11.0.0 (Windows 8.1 0.0.0) react-draggable props should render with defaultPosition set as string transform and handle subsequent translate() for DOM nodes FAILED
	AssertionError:   # specs\draggable.spec.jsx:255
	  
	  assert(style.indexOf('translate(calc(10% + 100px), calc(10% + 100px));') >= 0)
	         |     |                                                           |    
	         |     -1                                                          false
	         "touch-action: none;"                                                  
	  
	   at fail (specs/main.js:27336:3)
	   at ok (specs/main.js:27356:15)
	   at Decorator.prototype._callFunc (specs/main.js:30707:9)
	   at Decorator.prototype.concreteAssert (specs/main.js:30700:5)
	   at decoratedAssert (specs/main.js:31011:13)
	   at powerAssert (specs/main.js:4743:13)
	   at Anonymous function (specs/main.js:6344:7)

tnrich:
hmm looks like it is related to this: 
image

I'm not sure if there is some way around this..?

tnrich:
Okay found a solution based on this stackoverflow answer: https://stackoverflow.com/questions/21142923/ie-10-11-css-transitions-with-calc-do-not-work

this commit fixes it and tests should now pass :) 
https://github.com/mzabriskie/react-draggable/pull/361/commits/cf6979a81a3cd717dc464170fca0b1d3afbdf614


>updating createCSSTransform and createSVGTransform to use two translates instead of a calc function do to IE incompatibility; updating spec to reflect changes

tnrich:
@mzabriskie @STRML any thoughts?

tnrich:
bumping. @STRML is this project still active?

tnrich:
@STRML bump

lib/utils/domFns.js
```diff
@@ -109,13 +109,23 @@ export function offsetXYFromParent(evt: {clientX: number, clientY: number}, offs
   return {x, y};
 }
 
-export function createCSSTransform({x, y}: {x: number, y: number}): Object {
+export function createCSSTransform({x, y}: {x: number, y: number}, defaultPosition: {x: number | string, y: number | string }): Object {
```
STRML: You create a `DefaultControlPosition` type above but then don't use it here.

lib/utils/domFns.js
```diff
@@ -109,13 +109,23 @@ export function offsetXYFromParent(evt: {clientX: number, clientY: number}, offs
...
+  const defaultX = (typeof defaultPosition.x === 'string') ? defaultPosition.x : defaultPosition.x + 'px';
+  const defaultY = (typeof defaultPosition.y === 'string') ? defaultPosition.y : defaultPosition.y + 'px';
+  const translation = (defaultPosition && defaultPosition.x !== 0 && defaultPosition.y !== 0)
+  ? `translate(${defaultX}, ${defaultY}) translate(${x}px,${y}px)`
+  : 'translate(' + x + 'px,' + y + 'px)';
   // Replace unitless items with px
```
STRML: This comment no longer makes sense with the changes above.

lib/utils/domFns.js
```diff
@@ -109,13 +109,23 @@ export function offsetXYFromParent(evt: {clientX: number, clientY: number}, offs
   return {x, y};
 }
 
-export function createCSSTransform({x, y}: {x: number, y: number}): Object {
+export function createCSSTransform({x, y}: {x: number, y: number}, defaultPosition: {x: number | string, y: number | string }): Object {
+  const defaultX = (typeof defaultPosition.x === 'string') ? defaultPosition.x : defaultPosition.x + 'px';
```
STRML: This is really difficult to read. You create these two consts, `defaultX` and `defaultY`, but they're only actually used if and only if a `defaultPosition` is actually present. 

You then use that just to prepend what is being returned as a second translate. 

This is a lot of logic in what was previously a simple function and should be refactored such that the caller prepends the default translate if needed.

lib/utils/domFns.js
```diff
@@ -109,13 +109,23 @@ export function offsetXYFromParent(evt: {clientX: number, clientY: number}, offs
...
+  const defaultY = (typeof defaultPosition.y === 'string') ? defaultPosition.y : defaultPosition.y + 'px';
+  const translation = (defaultPosition && defaultPosition.x !== 0 && defaultPosition.y !== 0)
+  ? `translate(${defaultX}, ${defaultY}) translate(${x}px,${y}px)`
+  : 'translate(' + x + 'px,' + y + 'px)';
```
STRML: Why no template string here?

lib/utils/domFns.js
```diff
@@ -109,13 +109,23 @@ export function offsetXYFromParent(evt: {clientX: number, clientY: number}, offs
...
+  return {[browserPrefixToKey('transform', browserPrefix)]: translation };
 }
 
-export function createSVGTransform({x, y}: {x: number, y: number}): string {
-  return 'translate(' + x + ',' + y + ')';
+export function createSVGTransform({x, y}: {x: number, y: number}, defaultPosition: {x: number | string, y: number | string}): string {
+  const defaultX = (typeof defaultPosition.x === 'string') ? defaultPosition.x : defaultPosition.x + 'px';
+  const defaultY = (typeof defaultPosition.y === 'string') ? defaultPosition.y : defaultPosition.y + 'px';
+  const translation = (defaultPosition && defaultPosition.x !== 0 && defaultPosition.y !== 0)
+  ? `translate(${defaultX}, ${defaultY}) translate(${x}px,${y}px)`
```
STRML: This is wrong, svg transforms shouldn't have `px`.

STRML (changes requested):

lib/utils/domFns.js
tnrich: This should no longer have `px` in it

lib/utils/domFns.js
tnrich: Added template strings everywhere I could

lib/utils/domFns.js
tnrich: Reworked this to hopefully be more clear

lib/utils/domFns.js
tnrich: Removed the comment

lib/utils/domFns.js
tnrich: I now use it here

tnrich:
@STRML let me know if there is anything else I can change to get this PR accepted and merged? Thanks!

pavog:
Any updates on this? @tnrich @STRML

tnrich:
@pavog I believe I made all the changes @STRML requested the first time around. We're just waiting on him?

pavog:
@tnrich Yep, I believe that too. @STRML is still active (see #389) in this repository but seems to not want to review your changes again.

STRML (approved):

pavog:
@STRML Thank you very much 👍

eric-burel:
Awesome, thanks all for finishing this PR and merging !

tnrich:
@eric-burel @STRML had to revert the changes due to unexpected breakages. I've made a new PR that hopefully solves those issues by introducing a new property `initialPosition`  

https://github.com/mzabriskie/react-draggable/pull/393

PR #387, opened 2019-02-13
Allow rotation to pass through

jbborre:
Updated the transform css and svg attribute to allow for rotation to be passed through.


STRML:
Not sure I understand the use case here. I recommend double-wrapping your component instead of requiring a core lib modification.

cjshearer:
> Not sure I understand the use case here. I recommend double-wrapping your component instead of requiring a core lib modification.

Simple. Double wrapping breaks any `<Draggable` attempting to use `bounds="parent"`.

STRML:
You can use a querySelector for `bounds` as well.

That said, I can't accept this with all the `dist` changes.

OT-mliang:
Double wrapping also drags in relation to the rotated parent

STRML:
Closing this PR as stale. This PR has been open since 2019 without being merged. See PR #751 for a more recent implementation of rotation support. Thank you for the contribution!

Issue #315, opened 2018-02-19
When Draggable unmounts, if a textbox is focused, the textbox loses focus inexplicably


lukeknep:
My app renders some Items depending on the content of several text input.

The Items each contain a Draggable.

I'm coming across this very strange behavior:
- I type a valid value into a textbox, causing some Items to render
- Without leaving the textbox, I delete or add a few characters to make the value invalid
- The Items disappear (unmount)
- My textbox suddenly loses focus. However, no `onBlur` handlers are called, and if I switch tabs / windows and then switch back, focus is restored
- I can prevent this issue by deleting the Draggable component

Here's a codesandbox showing the issue:
https://codesandbox.io/s/32x3voojm5

I can consistently reproduce this.

Browser is Chrome 64.0.3282.119 (Official Build) (64-bit) on Mac OSX.

lukeknep:
Temporary workaround is to only render the Draggable when the component I care about is hovered, or only when the textbox is not focused. But both of those have flaws.

BTW this happens so consistently in our application that, without this workaround, we would not be able to use React Draggable in our application.

skvale:
Looks like it is an issue in the `enableUserSelectHack` prop

enableUserSelectHack causes this to be called
```javascript
export function removeUserSelectStyles(doc: ?Document) {
  try {
    if (doc && doc.body) removeClassName(doc.body, 'react-draggable-transparent-selection');
    window.getSelection().removeAllRanges();  // remove selection caused by scroll
  } catch (e) {
    // probably IE
  }
}
```

You could try
```javascript
<Draggable enableUserSelectHack={false}>
```

But I didn't look into what consequence not calling `removeUserSelectStyles` would have.

Working: https://codesandbox.io/s/944j9x101o

STRML:
Hmm. The user-select hack is necessary to ensure we don't start dragging text all over the page when using the draggable, but calling it on `componentWillUnmount` is a bit of a hack that ensures that we clean up after ourselves. It may be safe to completely remove it, or at the very least check if a drag is still in progress and only then clean up. But I believe `handleDragStop()` should be called in all circumstances.

RomaricMourgues:
I resolved it by adding :
```
enableUserSelectHack={false}
onStart={(event) => {$("body").addClass("dragging");}}
onStop={(event) => {$("body").removeClass("dragging");}}
```
on <Draggable /> and
```
body.dragging {
  -webkit-touch-callout: none;
  -webkit-user-select: none;
[...]

PR #388, opened 2019-02-25
Fix selection styles

paul-pro:
See https://github.com/mzabriskie/react-draggable/issues/315#issuecomment-459666957


webberwang:
Can we merge this? I'm using Antd & also faced the `::selection` bug, was very time consuming to hunt down this bug.

STRML:
Merged in 9998c562dee9aa9ab8aa2e5816ea0530ffadd105 and released in v3.3.2.

PR #392, opened 2019-03-05
Bounds change

torok-annamari95:
The react-draggable component has two issues when the "bounds" prop is used:

1. If the boundaries are changed in props on re-rendering, the new boundaries are not taken into account.
2. If the bounds are set to another HTML element, the resizing of the respective element is not taken into account. The most prominent example for this is that if a draggable bounds is set to html/body, resizing of the browser window makes the draggable sticking out.

There are no appropriate workarounds to these issues. 

travisoneill and wootencl submitted fixes for these. We, a small team of software professionals, merged these fixes with the current master, corrected the linter errors (that came due to a @flow issue), and propose these changes to be submitted.


timoxley:
I have a similar issue. Bounds are not respected if window is resized and there doesn't seem to be any way to force a bounds check without the user initiating a drag.

STRML:
Closing this PR as stale. This PR has been open since 2019 without being merged. If responding to bounds changes is still desired, please open a fresh PR against the current codebase. Thank you for the contribution!

Issue #391, opened 2019-03-01
3.2.0 Breaking Change: Drag Callbacks


STRML:
It appears there was an inadvertent breaking change in 3.2.0 relating to https://github.com/mzabriskie/react-draggable/pull/361.

Prior to 3.2.0, if a defaultPosition were present, it became the `x` and `y` present in the callback on the initial drag and subsequent to that.

In 3.2.0, the `x` and `y` received in the callback is _relative_ to the defaultPosition already passed. So if you pass a `defaultPosition` of `x: 100, y:100`and initiate`onDragStart`, the callback will contain `{x: 0, y: 0}`in the data. This makes sense if you consider`defaultPosition`to be something more like`initialOffset`, and is required if your offset is something like, for example, `{x: '10%', y: '10%'}`.

Ref: https://github.com/bokuweb/react-rnd/issues/437 and a broken build at https://circleci.com/gh/bokuweb/react-rnd/1340.

### Next Steps

Our choices:
- Is there a way to keep the old "absolute" offsets in callbacks from <= 3.1.1?
  - Percentage offsets will make this very difficult
- Release a major version instead?
- Drop the feature? A percentage-based default position can also be achieved by wrapping the `<Draggable>` in a div with a `transform: translate(10%, 10%)`.

Ping @tnrich.

**I've pulled 3.2.0 from npm** and will be releasing a 3.2.1 with just the TS export fix.

tnrich:
A couple ideas @STRML to "keep the old "absolute" offsets in callbacks from <= 3.1.1?": 


 - It seems like we could check to see if a percentage is being passed and only then do something more like an initial offset (this might be the most seamless as it would already work with existing packages)

 - could we introduce a new param, defaultPercentagePosition (or initialPosition). This would allow us to keep the current defaultPosition functionality the way it is. Only if you added a new property initialPosition would you get what the new functionality


If neither of the above ideas seem like viable solutions, then I think a major version would be most useful to my needs.

STRML:
I think a new param works, perhaps `positionOffset` - makes it clear that you're not pre-setting the `state` to something like `{x: 100, y: 100}`, you're offsetting the whole thing by some number or percentage and it doesn't actually affect the internal state and doesn't show up in callbacks.

tnrich:
@strml I think that is a great plan! Do you need help implementing it?

STRML:
Yeah, that would be great if you have the time!

tnrich:
[...]

PR #393, opened 2019-03-08
adding positionOffset prop to <Draggable/>

tnrich:
Hey @STRML, here is my first pass at allowing initial position to be a string percent. It is essentially using the same code as before but does not modify the defaultPosition prop in any way and instead adds a new initialPosition prop.

I think this is working as is. We should probably add a few more tests to be sure that the drag callbacks that broke last time aren't breaking this time.

connects https://github.com/mzabriskie/react-draggable/issues/391


tnrich:
Whoops.. I just realized that you wanted the new prop to be called `positionOffset` I'll change that now from what I had it as `initialPosition`

tnrich:
@STRML bump

lib/Draggable.js
```diff
@@ -178,6 +183,8 @@ export default class Draggable extends React.Component<DraggableProps, Draggable
       dragged: false,
 
       // Current transform x and y.
+      // x: props.position ? props.position.x : 0,
```
STRML: Comments should be removed

lib/utils/domFns.js
```diff
@@ -109,13 +109,28 @@ export function offsetXYFromParent(evt: {clientX: number, clientY: number}, offs
   return {x, y};
 }
 
-export function createCSSTransform({x, y}: {x: number, y: number}): Object {
-  // Replace unitless items with px
-  return {[browserPrefixToKey('transform', browserPrefix)]: 'translate(' + x + 'px,' + y + 'px)'};
+export function createCSSTransform({x, y}: ControlPosition, positionOffset: PositionOffsetControlPosition): Object {
+  let translation;
+  if (positionOffset && (positionOffset.x !== 0 || positionOffset.y !== 0)) {
```
STRML: Any reason to check for 0 here - it will still work w/o it and make the code a bit simpler.

lib/utils/domFns.js
```diff
@@ -109,13 +109,28 @@ export function offsetXYFromParent(evt: {clientX: number, clientY: number}, offs
   return {x, y};
 }
 
-export function createCSSTransform({x, y}: {x: number, y: number}): Object {
-  // Replace unitless items with px
-  return {[browserPrefixToKey('transform', browserPrefix)]: 'translate(' + x + 'px,' + y + 'px)'};
+export function createCSSTransform({x, y}: ControlPosition, positionOffset: PositionOffsetControlPosition): Object {
+  let translation;
```
STRML:
```suggestion
  let translation = `translate(${x}px,${y}px)`;
```

STRML: Could just define it here and prepend the default if present to save a few lines.

lib/utils/domFns.js
```diff
@@ -109,13 +109,28 @@ export function offsetXYFromParent(evt: {clientX: number, clientY: number}, offs
...
+    translation = `translate(${defaultX}, ${defaultY}) translate(${x}px,${y}px)`;
+  } else {
+    translation = `translate(${x}px,${y}px)`;
+  }
+  return {[browserPrefixToKey('transform', browserPrefix)]: translation };
 }
 
-export function createSVGTransform({x, y}: {x: number, y: number}): string {
-  return 'translate(' + x + ',' + y + ')';
+export function createSVGTransform({x, y}: ControlPosition, positionOffset: PositionOffsetControlPosition): string {
```
STRML: Really feels like this could be merged w/ the previous function, perhaps with a param (default `'px'`) to determine translate unit suffix.

STRML (changes requested):
Good stuff overall, just some code style suggestions to make this simpler.

lib/utils/domFns.js
tnrich: I was thinking we would need the transform if 0 was passed, but I guess in that case the transform doesn't do anything. I'll remove it.

lib/utils/domFns.js
tnrich: Sounds good

lib/utils/domFns.js
tnrich: Done

lib/utils/domFns.js
tnrich: Done

tnrich:
@STRML I've made the changes you requested. Let me know if those check out!

tnrich:
I was having trouble getting `npm install` to run without errors (npm v6.9) so I pushed the latest commits without running the precommit hooks (just FYI)

STRML:
What were the errors?

tnrich:
@STRML hmm it just worked with `yarn install` but was breaking with `npm install`. I can't reproduce what I was seeing now so I guess just disregard that.

tnrich:
@STRML merge it?

lib/utils/domFns.js
```diff
@@ -109,13 +109,23 @@ export function offsetXYFromParent(evt: {clientX: number, clientY: number}, offs
...
+  return {[browserPrefixToKey('transform', browserPrefix)]: translation };
 }
 
-export function createSVGTransform({x, y}: {x: number, y: number}): string {
-  return 'translate(' + x + ',' + y + ')';
+export function createSVGTransform(controlPos: ControlPosition, positionOffset: PositionOffsetControlPosition): string {
+  const translation = getTranslation(controlPos, positionOffset, '');
+  return translation;
+}
+export function getTranslation({x, y}: ControlPosition, positionOffset: PositionOffsetControlPosition, unitSuffix: string): string {
```
STRML, line 121: 👍

lib/Draggable.js
```diff
@@ -121,6 +122,10 @@ export default class Draggable extends React.Component<DraggableProps, Draggable
       x: PropTypes.number,
       y: PropTypes.number
     }),
+    positionOffset: PropTypes.shape({
```
STRML: This should be a maybe type

STRML (approved):

tnrich:
Lemme know when this gets released! Thanks @STRML

tnrich:
@STRML will you please release a new version with this in it? Thank you :)

STRML:
Published as `3.3.0` @tnrich.

tnrich:
Thanks!!

Issue #363, opened 2018-09-14
update element position automatically when bounds is updated


o02112:
when bounds changed

For example "bounds" area becomes smaller.
Update draggable element position automatically, in case of  element getting out of the visible area.
Or just trigger the onDrag event once.

JSerhii:
@o02112 I have the same problem. Did you find solution for this bug?

EitanElbaz:
Having the same problem. Has a solution to this been found yet?

EitanElbaz:
I've solved this problem by essentially faking a mouse down event. You'll have to decide for yourself when to call this `fakeDrag` function obviously, but I've included a window resize example. Tested on Chrome 72 and version 3.2.1 of `react-draggable`.

```javascript
class SomeReactComponent extends React.Component<Props> {

    private childRef: React.RefObject<HTMLDivElement>;

    constructor(props) {
        super(props);
        this.childRef = React.createRef();
    }

    componentDidMount() {
        window.addEventListener('resize', this.onWindowResize);
    }

    componentWillUnmount() {
        window.removeEventListener('resize', this.onWindowResize);
    }

    render() {
        const props = this.props;

        return (
            <Draggable
                axis={"x"}
                bounds={"parent"}
            >
                <div
                    ref={this.childRef}
                >
                    {props.children}
                </div>
            </Draggable>
        )
    }

    private onWindowResize = () => {
        this.fakeDrag();
    };

    private fakeDrag = () => {
        const clickEvent = document.createEvent ('MouseEvents');
        clickEvent.initEvent ("mousedown", true, true);
        this.childRef.current.dispatchEvent(clickEvent);
    }
}
```

tnrich:
I wasn't able to get @EitanElbaz's solution to work for me but I made some modifications and now it is working. The key was triggering a mousemove on the document: 

```js

  onWindowResize = () => {
    //trigger a fake drag event here so that the dialog will stay in the window on window resize
    const targetNode = document.querySelector(`.${Classes.DIALOG_HEADER}`);
    if (targetNode) {
      //--- Simulate a natural mouse-click sequence.
      triggerMouseEvent(targetNode, "mouseover");
      triggerMouseEvent(targetNode, "mousedown");
      triggerMouseEvent(document, "mousemove");
      triggerMouseEvent(targetNode, "mouseup");
      triggerMouseEvent(targetNode, "click");
    } else console.log("*** Target node not found!");

    function triggerMouseEvent(node, eventType) {
[...]

PR #397, opened 2019-03-30
Observable prop that will trigger position recalculation when changed 

Lowhearth:
Draggable item had no way to identify when the bounds element changed.  Fixes#363 Fixes#270

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,11 @@ axis: string,
 //   can be moved.
 bounds: {left: number, top: number, right: number, bottom: number} | string,
 
+// Observable value that in the case of change, will recalculate the position
+// of the Draggable element, (useful for keeping the element in bounds after
+// resizing) 
+triggerBounds: any,
+
 // Specifies a selector to be used to prevent drag initialization.
 // Example: '.body'
 cancel: string,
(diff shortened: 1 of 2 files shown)


STRML:
Why introduce a new prop when we can just watch `bounds`?

Lowhearth:
Attempts to watch the prop bounds (see #363 ) as been made but involved a lot of code for such a small feature. I'll list the reasons why I've choosen this aproach:

1. Bounds are tricky to watch since there is no recordings of the previous bounds at the state level (I'm talking about the actual bounds and not the prop, since we are accepting a string as 'parent').

2. All this bounds comparation will add extra steps to the componentwillreceiveprops method which may cause performance issues in the future.

If you prefer the watching the bounds prop approach I'll give it a try.

Thanks,

STRML:
I agree the other `bounds` features are pretty hard to swallow. This is a better solution in terms of simplicity, and would allow the dev - if he/she wishes - to set up observables and increment this counter.

STRML:
Could you add something to the README?

Lowhearth:
I've already added the new prop inside the draggable props section: 


> // Observable value that in the case of change, will recalculate the position
> // of the Draggable element, (useful for keeping the element in bounds after
> // resizing) 
> triggerBounds: any,


Any additional suggestions @STRML ?

PR #412, opened 2019-07-05
Prevent scrolling on mobile devices, like ipad/iphone.

### LINKED ISSUE #406: The whole page is dragged on iOS browsers
See the demo page's behavior on Safari(iOS12)

ezgif com-video-to-gif

Other browsers with the same problem:
- Chrome Mobile(iOS)
- Edge Mobile(iOS)
- Firefox Mobile(iOS)

It seems good on Android.

GilHyeon:
fix #406 

> When dragging in Safari on iDevices (I specifically tested an iPhone 7 and iPad Air 2, both running iOS 10.3) the entire screen scrolls with the drag

It seemed to solve this problem in the past, but it did not work properly.

I add some code. it works.

### DIFF
--- a/lib/DraggableCore.js
+++ b/lib/DraggableCore.js
@@ -221,6 +221,14 @@ export default class DraggableCore extends React.Component<DraggableCoreProps, D
     touchIdentifier: null
   };
 
+  componentDidMount() {
+    const thisNode = ReactDOM.findDOMNode(this);
+    if (thisNode) {
+      // Prevent scrolling on mobile devices, like ipad/iphone.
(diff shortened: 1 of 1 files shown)


niculistana:
bumping this fix. Also it may be a good idea to handle dispatch/bubble-up a `mousemove` event when `disabled=true` to allow scrolling when a `mousemove` event is fired on a disabled draggable component.

GilHyeon:
I checked the recent demo on the iPad and it seems that this problem is solved. Can you tell when it's resolved? I'll close this PR.

PR #416, opened 2019-07-25
add head element if no head element exists for styles

wzsun:
In certain cases there exists no head elements for example when you view a pdf file in a browser. This  creates a head element if none exists.


STRML:
I don't think this belongs in the lib. If you are handling pages where there is no `<head>` element you are going to run into problems with tons of libraries. Insert one in your application instead.

wzsun:
Hey, thanks didn't think about that.

PR #418, opened 2019-07-29
Change link in readme to example source

panzarino:
It's far more useful to have the link go to the JS code that was used to create the page rather than the HTML layout of the page


STRML:
Thanks!

PR #425, opened 2019-09-07
fix: upload cjs to npm instead of es2015

### LINKED ISSUE #426: Cannot Import
it requires ./build/draggable, but there is no build folder when I import after installing from node, only lib. As a result nothing is exported

orzyyyy:
My test units crash as below:
```jsx
/home/orzorzorzorz/github/mini-xmind/node_modules/react-draggable/lib/Draggable.js:2
import React from 'react';
       ^^^^^

SyntaxError: Unexpected identifier
```

Close #426

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "flow": "flow"
   },
   "files": [
-    "/lib"
+    "build"
   ],
   "typings": "./typings/index.d.ts",
   "types": "./typings/index.d.ts",


orzyyyy:
Check [here](https://circleci.com/gh/orzyyyy/mini-xmind/2872?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) for detail.

What upload to npm is the source code as `es2015`, which contains `import`. That's why my test units crash.

PR #430, opened 2019-09-15
Migrate from findDomNode to refs

senadir:
this is an initial PR to migrate from the now deprecated `findDomNode` to using `ref`

functionality wise this PR seems to work, there are however some problems that I could use some help in:

- the flow types are not correct, I've never used flow before so I'm not sure how to fix and didn't have much time to research for it.
- I couldn't run the tests from some reason (they run immediately but nothing happens).
- I would rather refactor this to use `forwardRef` and use the same ref from `Draggable` into `DraggableCore` but it didn't work for some reason, but alas, the current approach still seems to work

### DIFF
--- a/lib/Draggable.js
+++ b/lib/Draggable.js
@@ -31,6 +31,7 @@ export type DraggableProps = {
   defaultPosition: ControlPosition,
   positionOffset: PositionOffsetControlPosition,
   position: ControlPosition,
+  draggableRef: Element,
   scale: number
 };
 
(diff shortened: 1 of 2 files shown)


STRML:
This is actually quite difficult to do well, because wrapped elements *must* properly forward refs down to the underlying DOM node, which is a very breaking change and will require significant modification on many apps that do something like:

```
<Draggable>
  <Panel>
    <Panel.Title>...</Panel.Title>
    <Panel.Body>...</Panel.Body>
  </Panel>
</Draggable>
```

I expect components to exert this kind of hygiene into React 17, but there will be some hard-to-solve edge cases.

vkrol:
@STRML FYI https://github.com/reactjs/react-transition-group/blob/master/CHANGELOG.md#440-2020-05-05

STRML:
Interesting idea! I'd be happy to accept a PR with this.

vkrol:
@STRML If I understand correctly, this PR contains a similar solution.

STRML:
Unfortunately, it doesn't. The `draggableRef` and `elementRef` (why different names?) are ignored if passed in `props`.

thienandangthanh:
#478 supersedes this PR.
I think this PR should be closed.

PR #431, opened 2019-09-17
Use container offset in bounds computation

florianduros:
Hello,

When the bounds are provided as html selector, their computations doesn't use the container position. So the draggable can move in the selector area. The area size is the same than the selector moreover the position is different because the bounds don't use the selector offset.

I made the required changes and added a test.

### DIFF
--- a/lib/utils/positionFns.js
+++ b/lib/utils/positionFns.js
@@ -32,12 +32,10 @@ export function getBoundPosition(draggable: Draggable, x: number, y: number): [n
     const boundNodeStyle = ownerWindow.getComputedStyle(boundNode);
     // Compute bounds. This is a pain with padding and offsets but this gets it exactly right.
     bounds = {
-      left: -node.offsetLeft + int(boundNodeStyle.paddingLeft) + int(nodeStyle.marginLeft),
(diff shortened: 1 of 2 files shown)


lamflam:
I just spent a few hours tracking down why I couldn't my my div to drag partially offscreen to the left with a `top: -200px` on the bounding parent and tracked it down to this same thing. Really wish I had seen this PR before spending that time. Would love to see this get merged!

STRML:
This depends on whether the containing element uses `position: relative` or not, yes? If so, there should be no need to use `boundNode.offset{Left,Top}` because the draggable's absolute position is relative to the container.

florianduros:
Yes if the container is a parent of the draggable element.

In my case of use, the draggable element is not a child of the html element pointed by the bounds.

STRML:
It seems that we should perhaps accept a function as bounds, returning coords.

dhkatz:
I can confirm that this is basically needed to work properly with CSS grids as well it seems.

Without this, react-draggable seems to think that the bounds of the parent were where the div was before being positioned by `grid-row` and `grid-column` on other divs within the grid because it causes the div with the draggables to be shifted around.

PR #445, opened 2019-12-04
Add transform-inline-environment-variables Babel plugin

erictooth:
The cjs and esm builds introduced 4.1.0 contain references to **process.env.DRAGGABLE_DEBUG**, which is not necessarily replaced by all bundlers out of the box.

My suggestion would be to do the same thing as your Webpack configuration, and rewrite process variables at build time using [babel-plugin-transform-inline-environment-variables](https://babeljs.io/docs/en/babel-plugin-transform-inline-environment-variables).


STRML:
Fair enough, thanks for the PR.

PR #446, opened 2019-12-16
Expand .cancel documentation

### LINKED ISSUE #436: Allow `cancel` option to be given as Array
`cancel` currently only allows a single selector but in complex use cases, it would be handy to specifiy multiple CSS selectors to allow fine-grained control over the elements that don't start the dragging.

This feature could be made backwards-compatible by using a `Array.isArray` check so both `string` and `Array` types could be supported.

silverwind:
Fixes: https://github.com/mzabriskie/react-draggable/issues/436

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,7 +178,8 @@ axis: string,
 //   can be moved.
 bounds: {left: number, top: number, right: number, bottom: number} | string,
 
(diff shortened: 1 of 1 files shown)


STRML:
Thanks!

PR #447, opened 2019-12-28
fix(user-select): move addUserSelectStyles from handleDragStart to ha…

fxxjdedd:
If it's just a simple click, don't modify the DOM.


STRML:
This will add the styles multiple times, I doubt that is desired.

fxxjdedd:
> This will add the styles multiple times, I doubt that is desired.

This is indeed a problem, Maybe we need an extra `if` in `handledrag`.

However, it is necessary not to add styles when clicking. Because in actual development, we do have two different types of logic: click and drag. When there are too many heavy `GridItem`s, click can cause some performance problems

fxxjdedd:
@STRML 
Hello,
I found that `addClassName` already has logic to avoid duplicate classes. However, I can add another `hasUserSelectStyles` judgment if necessary.

STRML:
Closing this PR as stale. This PR has been open since 2019 without being merged. If this user-select timing issue persists, please open a fresh PR against the current codebase. Thank you for the contribution!

PR #464, opened 2020-04-01
Add rotation prop

bluematter:
I'm adding rotation to objects in https://storycreatorapp.com/ and think it would be nice to have the ability to pass in rotation to the component.

It can be a pain placing the rotation in the parent component. Passing it through directly works quite well for my needs and I assume other folks needs.

### DIFF
--- a/lib/Draggable.js
+++ b/lib/Draggable.js
@@ -31,7 +31,8 @@ export type DraggableProps = {
   defaultPosition: ControlPosition,
   positionOffset: PositionOffsetControlPosition,
   position: ControlPosition,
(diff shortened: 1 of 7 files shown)


STRML:
Hi. I don't know what this really does, do you have a demo?

Please don't also adjust the whole project's styling in a PR. Your preferences are not the project's preferences and it obfuscates what you are actually changing.

Additionally it is not possible to accept a PR that has a version bump in it.

STRML:
Going to close this as a dupe of #387 until there's a more clear PR & justification.

Issue #400, opened 2019-04-24
Can DraggableBounds properties be optional


thomaskbird:
Can DraggableBounds `left`, `right`, `top` and `bottom` be optional? The use case is that I have set `axis={"x"}` and I would like to limit the dragging to:
```
bounds={{
   left: -190,
   right: 0
}}
```
but I can't because the interface is currently:
```
export interface DraggableBounds {
    left: number
    right: number
    top: number
    bottom: number
}
```
and I would like it to be:
```
export interface DraggableBounds {
    left?: number
    right?: number
    top?: number
    bottom?: number
}
```

lwolle:
Hey there, Is there an answer to this question. Does `ReactDraggable` expect all props to be passed in, or is it just coincidence that it doesn't throw when some properties of `bounds` missing.

grhallenbeck:
The documentation indicates these are optional arguments, the type definitions should reflect that.
```
     * Alternatively, pass an object with the following properties, all of which are optional:
     *
     * {left: LEFT_BOUND, right: RIGHT_BOUND, bottom: BOTTOM_BOUND, top: TOP_BOUND}
```

STRML:
Fixed in #473 for both TS & Flow definitions.

PR #473, opened 2020-04-14
fix: make DraggableBounds properties optional in typings

grhallenbeck:
Per #400, the typings for DraggableBounds should allow the properties to be optional.

This updates the types for DraggableBounds to match the documentation and the PropType definitions.


STRML:
Thanks!

Issue #465, opened 2020-04-02
Warning: findDOMNode is deprecated in StrictMode.


bharadhwaj:
I am getting an error when I am trying to use Draggable,

> Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of Draggable which is inside StrictMode. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here: https://fb.me/react-strict-mode-find-node

Seems like an error inside the library code for me, and not very sure is there anything I can do to solve the error in my end?

I am using uncontrolled Draggable component. 

Any help is appreciated.

ricardo-jsx:
I believe that this comment may help you: https://github.com/raatmarien/playcardswith.me/issues/56#issuecomment-608322650

TL;DR: This warning comes from `<React.StrictMode />` and, for now, it hasn't caused any unexpected behavior and the only way to remove it is to remove from your project `React.StricMode />` itself.

STRML:
Duplicate of #440. Accepting PRs but I don't have the time to fix this properly yet.

PR #478, opened 2020-05-11
feat(React): add `nodeRef` prop

### LINKED ISSUE #440: Support `<React.StrictMode>`
When using `<React.StrictMode>` I receive the following browser console error:

```
Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of Draggable which is inside StrictMode. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here: https://fb.me/react-strict-mode-find-node
```

STRML:
If running in React Strict mode, ReactDOM.findDOMNode() is deprecated.
Unfortunately, in order for <Draggable> to work properly, we need raw access
to the underlying DOM node. If you want to avoid the warning, pass a `nodeRef`
as in this example:

```js
function MyComponent() {
  const nodeRef = React.useRef(null);
  return (
    <Draggable nodeRef={nodeRef}>
      <div ref={nodeRef}>Example Target</div>
    </Draggable>
  );
}
```

This can be used for arbitrarily nested components, so long as the ref ends up
pointing to the actual child DOM node and not a custom component.

Thanks to react-transition-group for the inspiration.
`nodeRef` is also available on <DraggableCore>.

Supersedes #430 
Fixes #440, #465

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,28 @@ onDrag: DraggableEventHandler,
 // Called when dragging stops.
 onStop: DraggableEventHandler,
 
+// If running in React Strict mode, ReactDOM.findDOMNode() is deprecated.
+// Unfortunately, in order for <Draggable> to work properly, we need raw access
+// to the underlying DOM node. If you want to avoid the warning, pass a `nodeRef`
+// as in this example:
+//
+// function MyComponent() {
+//   const nodeRef = React.useRef(null);
+//   return (
+//     <Draggable nodeRef={nodeRef}>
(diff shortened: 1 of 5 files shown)


dimazuien:
Looks like nodeRef is missing in typings

Jack-Works:
> Looks like nodeRef is missing in typings

Does it added now?

dimazuien:
Nope, I had this issue on version 4.4.2 which is still the latest one

STRML:
Released in v4.4.3

PR #498, opened 2020-06-27
Consider ctrl-click as non-left-click

assembled-michael:
We noticed that the behavior on Draggable was different when right-clicking vs. ctrl-clicking, even though they should be treated the same. This modifies it so ctrl-clicks are not considered as left-clicks.

### DIFF
--- a/lib/DraggableCore.js
+++ b/lib/DraggableCore.js
@@ -274,7 +274,12 @@ export default class DraggableCore extends React.Component<DraggableCoreProps, D
     this.props.onMouseDown(e);
 
     // Only accept left-clicks.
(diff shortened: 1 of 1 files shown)


misumi-khoi-nguyen-01 (approved):

misumi-khoi-nguyen-01:
Good

STRML:
Closing this PR as stale. This PR has been open since 2020 without being merged. If ctrl-click handling is still an issue, please open a fresh PR against the current codebase. Thank you for the contribution!

STRML:
Reopening - this is a correct fix for macOS ctrl+click behavior.

STRML:
Closing in favor of #786 which implements this fix against the current codebase. Thank you for the contribution!

PR #507, opened 2020-07-29
fix(grid): incorrect `x` and `y` passed to onStop

### LINKED ISSUE #413: Bug: onStop returns wrong positions when using grid
When using grid, the x and y returned by onStop are incorrect. lastX and lastY however seem to be correct. I can't find a explanation for the values onStop returns for x and y.
I've created an example showing the behavior:

https://codesandbox.io/s/texthighlights-wtrzt

Please have a look at the console output when dragging the object around. I would have expected x and y to change based on the grid of [50,50] but this is not the case.

React-draggable is also used by react-rnd - they also have an issue open for this bug (https://github.com/bokuweb/react-rnd/issues/453)

STRML:
This was caused by `onStop` erroneously recalculating coordinates,
which is not necessary / should not happen in the real world, as `mouseup`
and `touchend` should not pass `clientX` and `clientY` values meaningfully
different than the last corresponding move event. In fact, they are already
ignored by `<Draggable>`.

For this reason, we can simply trust the `lastX` and `lastY` in `onDragStop`,
and pass those back.

Fixes #413, bokuweb/react-rnd#453

### DIFF
--- a/lib/DraggableCore.js
+++ b/lib/DraggableCore.js
@@ -381,11 +381,8 @@ export default class DraggableCore extends React.Component<DraggableCoreProps, D
 
   handleDragStop: EventHandler<MouseTouchEvent> = (e) => {
     if (!this.state.dragging) return;
-
-    const position = getControlPosition(e, this.state.touchIdentifier, this);
-    if (position == null) return;
-    const {x, y} = position;
-    const coreEvent = createCoreData(this, x, y);
+    // No dragging is possible on stop, so we just echo back lastX & lastY.
+    const coreEvent = createCoreData(this, this.state.lastX, this.state.lastY);
 
     // Call event handler
     const shouldContinue = this.props.onStop(e, coreEvent);
--- a/specs/draggable.spec.jsx
+++ b/specs/draggable.spec.jsx
@@ -153,6 +153,54 @@ describe('react-draggable', function () {
       simulateMovementFromTo(drag, 0, 0, 100, 100);
     });
 
+    // https://github.com/STRML/react-draggable/issues/413
+    it('should adjust draggable data output when `grid` prop supplied', function () {
+      let dragCount = 0;
+      function onDrag(event, data) {
+        if (dragCount === 0) {
+          // Should be 150, not 160
+          assert(data.x === 150);
+          assert(data.y === 150);
+          assert(data.lastX === 0);
+          assert(data.lastY === 0);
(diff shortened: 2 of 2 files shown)


STRML:
@nicholaschiang could you verify this works for you - simply pin `react-draggable` to this git branch in your package.json `resolutions`?

```json
"react-draggable": "git+https://git@github.com/strml/react-draggable.git#fix/grid-dragStop"
```

nicholaschiang:
Just added it @STRML and it looks like the issue is still there:

jumping

You'll notice how the draggable keeps jumping a couple of pixels after dropping (there are some other unrelated issues shown in that GIF too... feel free to ignore them until I finish debugging on my end).

nicholaschiang:
@STRML I also created a little repro of the same (or, at least, a similar) problem I was having when dynamically resizing the grid snap and RND size.

https://codesandbox.io/s/white-water-xrqp2?file=/src/index.js

You'll notice in that repro that the RND keeps jumping around and doesn't snap correctly to the resized grid (even though we're using the `{ x: data.lastX, y: data.lastY }` workaround proposed in #413):

resizing-parent

Ideally, the X position should always be a multiple of the RND's width (displayed on the parent) and the Y position should always be a multiple of the RND's height (also displayed on the parent).

trickpattyFH20:
Any idea on on a timeline for this? @STRML

trickpattyFH20:
> @nicholaschiang could you verify this works for you - simply pin `react-draggable` to this git branch in your package.json `resolutions`?
> 
> ```json
> "react-draggable": "git+https://git@github.com/strml/react-draggable.git#fix/grid-dragStop"
> ```

@STRML  The repository does not exist at this link anymore.
Also, when trying to use the `grid-dragStop` branch from this repo as an npm dependency, the source code for this repo is not included in `./node-modules` because the package.json in `react-grid-layout/react-draggable` does not include `./lib` in the  [files](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#files) field, or have`./build` committed to that branch.

Could you merge a fix for this issue, or include the source code on that branch so that it can be used as a dependency?

STRML:
Closing this PR as stale. While this grid snap fix was identified in 2020, the discussion shows there may be edge cases that weren't fully resolved. If this issue persists, please open a fresh issue or PR against the current codebase. Thank you\!

PR #525, opened 2020-10-09
Adding `react` and `react-dom` as dependencies for Yarn PnP

cozimacode:
Yarn V2 refuses access to packages which are not listed as a dependency. This should help fix it.

Details: https://yarnpkg.com/advanced/rulebook#packages-should-only-ever-require-what-they-formally-list-in-their-dependencies

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
     "prop-types": "^15.6.0"
   },
   "peerDependencies": {
-    "react": "^16.13.1",
-    "react-dom": "^16.13.1"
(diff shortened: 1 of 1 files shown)


STRML:
Should just be `dependencies` not `peerDependencies`, yeah?

cozimacode:
> 
> 
> Should just be `dependencies` not `peerDependencies`, yeah?

Nope. It should be `peerDependencies`. My bad. I forgot to change the title after my second commit :)

STRML:
Thanks!

PR #534, opened 2020-11-29
Draggable component freezes when content is changing dynamically.

jgeliberte:
The matchesSelectorAndParents to function throws error when the querySeelctor.all() is being called.
This only happens when the Draggable component uses handle / cancel.

### DIFF
--- a/lib/utils/domFns.js
+++ b/lib/utils/domFns.js
@@ -31,9 +31,14 @@ export function matchesSelector(el: Node, selector: string): boolean {
 export function matchesSelectorAndParentsTo(el: Node, selector: string, baseNode: Node): boolean {
   let node = el;
   do {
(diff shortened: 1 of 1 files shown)


STRML:
Can you show an example in CodeSandbox where the current version crashes? Also, could you add a test?

jgeliberte:
i could not replicate this bug on a fresh installed mui-datatables and react-draggable however. On my local setup, i checked all components affected by the draggable component and the mui-datatable is the culprit when changing the content of the datable.

This is the error shown when the datatable rows changed (changed to 25 to 50 rows per page).
image

And noticed that the domfns.js on function `matchesSelectorAndParentsTo` it does not handle null values. so by adding a try catch or handling the null value and returning it false fixed the error.

PR #541, opened 2021-02-10
Added drop-detection to examples

programstinator:
There have been a few repeated requests for drop-detection so I extended the example page with two additional boxes that use a simple conditional css class trick- I apply the `pointer-events: none` style to the box that is being dragged. This enables the developer to catch the appropriate event target, both when entering the drop target and releasing the mouse button. 
I haven't made any changes to the source code itself.

### DIFF
--- a/example/example.js
+++ b/example/example.js
@@ -28,6 +28,21 @@ class App extends React.Component {
   onStop = () => {
     this.setState({activeDrags: --this.state.activeDrags});
   };
+  onDrop = (e) => {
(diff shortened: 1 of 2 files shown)


fcalizobrandcoders:
Super helpful, thanks!

STRML:
Nice, thanks!

hugoburguete:
@programstinator Are you able to update the example github site? Currently it doesn't show it.

PR #545, opened 2021-03-05
Reconcile handling of mouse and touch drag start events

szakharchenko:
Attach the mousedown event handler to the node, likewise to the touchstart, so that it may be cancelled if needed and otherwise handled more uniformly

### DIFF
--- a/lib/DraggableCore.js
+++ b/lib/DraggableCore.js
@@ -243,6 +243,7 @@ export default class DraggableCore extends React.Component<DraggableCoreProps, D
     // https://developers.google.com/web/updates/2017/01/scrolling-intervention
     const thisNode = this.findDOMNode();
     if (thisNode) {
+      addEvent(thisNode, eventsFor.mouse.start, this.onMouseDown, {passive: false});
       addEvent(thisNode, eventsFor.touch.start, this.onTouchStart, {passive: false});
     }
   }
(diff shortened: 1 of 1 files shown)


STRML:
Hmm, this is interesting. Can you point me to a use case / example code where this change makes a difference for you?

szakharchenko:
I'm not sure I can. Having different code paths for mouse and touch events just looks inelegant (e.g. one receives a browser event and another a synthetic event). I made this change as part of handling click-vs-drag issues like https://github.com/react-grid-layout/react-draggable/issues/531, https://github.com/react-grid-layout/react-draggable/issues/419 and https://github.com/react-grid-layout/react-draggable/issues/409 . It seems like I've made things to work even without this change though, so I can't say it's absolutely necessary, it just looks like it would make things a bit cleaner. This is a merge offer, not a pull request;)

szakharchenko:
The comment would need to be reworded though; this doesn't by itself make it possible to cancel the click event after drag.

STRML:
That makes sense. I'll test it and see if this changes anything upstream.

STRML:
Closing this PR as stale. This PR has been open since 2021 without being merged. If mouse/touch event handling still needs improvement, please open a fresh PR against the current codebase. Thank you for the contribution!

PR #546, opened 2021-03-05
Fix git URL installation

szakharchenko:
These commits make it possible to install this package from a git URL.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "test-ie": "karma start --browsers=IE --single-run=false --auto-watch=true",
     "dev": "make dev",
     "build": "make clean build",
+    "prepare": "make clean build",
     "lint": "make lint",
     "flow": "flow"
   },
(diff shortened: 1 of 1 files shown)


STRML:
Thanks!

STRML:
Although, why avoid using make/yarn at prepare time? We want to ensure we're really using the correct build script, and this adds maintenance overhead if we change the build.

szakharchenko:
The answer to your direct question is: to allow installation from Git URLs in environments where either yarn or make are not installed; likely that's not what you wanted to hear.

I would ask the opposite question, i.e. 'why use make&yarn at all instead of having the command line mentioned in the commit go to the "build" script, and have "prepare" just run "build"?'. It's a matter of taste, and up to you really, I don't intend to force you in any way; you can always merge only the first patch, or ignore the issue entirely.

STRML:
Would agree with you that Make is a superfluous dependency here and isn't really used for anything that `scripts` can't already handle.

STRML:
Closing this PR as stale. This PR has been open since 2021 without being merged. If git URL installation is still an issue, please open a fresh PR against the current codebase. Thank you for the contribution!

PR #556, opened 2021-04-16
optimizations: setState in lifecycles + forced reflow

### LINKED ISSUE #554: A couple of small performance related issues
We're working on polishing our product performance-wise, and have identified a couple of small issues with this (awesome!) library that we'd like you guys to address. Luckily, both are pretty trivial to fix:

1.
image

This one originates here:
https://github.com/react-grid-layout/react-draggable/blob/master/lib/Draggable.js#L239

An unconditional setState in a lifecycle. Something like this would fix that:
```
componentWillUnmount() {
    if (this.state.dragging) {
        this.setState({dragging: false}); // prevents invariant if unmounted while dragging
    }
}
```

2.
Sometimes we get "forced reflow" alerts like these:

image

Those originates here:
https://github.com/react-grid-layout/react-draggable/blob/master/lib/DraggableCore.js#L261

Wrapping the DOM mutation in a rAF would fix that:
```
if (this.props.enableUserSelectHack) {
    requestAnimationFrame(() => {
        removeUserSelectStyles(ownerDocument);
    });
}
```

avoskresensky:
Closes #554

### DIFF
--- a/lib/Draggable.js
+++ b/lib/Draggable.js
@@ -236,7 +236,9 @@ class Draggable extends React.Component<DraggableProps, DraggableState> {
   }
 
   componentWillUnmount() {
-    this.setState({dragging: false}); // prevents invariant if unmounted while dragging
+    if (this.state.dragging) {
+      this.setState({dragging: false}); // prevents invariant if unmounted while dragging
+    }
   }
 
   // React Strict Mode compatibility: if `nodeRef` is passed, we will use it instead of trying to find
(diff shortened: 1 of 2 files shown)


andresgutgon:
Looks good to me

AngeloWangCoder:
@STRML Hi, thank you for the amazing library. Since this issue has been resolved, I hope that react-grid-layout can be migrated to a newer version of react-draggable.

PR #577, opened 2021-06-05
Adding ability to control state from refs during onDrag

bluematter:
**Problem**

When dragging it's a challenge to manage state from refs because internally `this.setState` will always override any `refs.setState` making it useless.

Even if you return false from the function DraggableCore will `this.handleDragStop(new MouseEvent('mouseup'));` trigger drag stop.

**Solution**

Return an object with a key `controlled` and if it's true it will short circuit onDrag giving the custom callback the ability to control the state.

https://www.loom.com/share/2ca4ef824bc2404bac0cbf2f8108de2c

### DIFF
--- a/lib/Draggable.js
+++ b/lib/Draggable.js
@@ -296,6 +296,8 @@ class Draggable extends React.Component<DraggableProps, DraggableState> {
 
     // Short-circuit if user's callback killed it.
     const shouldUpdate = this.props.onDrag(e, uiData);
+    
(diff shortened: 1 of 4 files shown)


STRML:
Closing this PR as stale. This PR has been open since 2021 without being merged. If controlling state from refs during onDrag is still desired, please open a fresh PR against the current codebase with updated implementation. Thank you for the contribution!

PR #599, opened 2021-08-26
Replace classnames with clsx

wojtekmaj:
clsx is fully compatible with classnames API, is quickly gaining traction and is about 1/3rd smaller than classnames.


STRML:
Thanks!

PR #621, opened 2022-01-07
`grid` prop unused in `handleDragStop`

### LINKED ISSUE #413: Bug: onStop returns wrong positions when using grid

BrianHung:
Related: https://github.com/react-grid-layout/react-draggable/pull/507

Fixes: https://github.com/react-grid-layout/react-draggable/issues/413

### DIFF
--- a/lib/DraggableCore.js
+++ b/lib/DraggableCore.js
@@ -389,7 +389,6 @@ export default class DraggableCore extends React.Component<DraggableCoreProps, D
     if (Array.isArray(this.props.grid)) {
       let deltaX = x - this.state.lastX, deltaY = y - this.state.lastY;
       [deltaX, deltaY] = snapToGrid(this.props.grid, deltaX, deltaY);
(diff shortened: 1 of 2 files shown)


STRML:
Thanks, would you mind adding a test please so this doesn't regress in the future? Thanks.

BrianHung:
@STRML Tell me if that works for you!

STRML:
Thanks!

PR #632, opened 2022-02-09
Better strict mode support

quolpr:
If nodeRef passed, but `ref.current` returns undefined it will fall back to the `ReactDOM.findDOMNode` and trigger react error that findDOMNode is deprecated in StrictMode. 

The new behavior makes sure that if nodeRef passed, then `ReactDOM.findDOMNode` will never be used, even if the ref value is undefined


STRML:
Thanks!

nikolas:
It looks like this change may be causing this issue many people are running into here: https://github.com/react-grid-layout/react-grid-layout/issues/1732

PR #635, opened 2022-02-20
Clarified how `bounds` should be used

duckboy81:
Added a clue for future users indicating 'how far in each direction` does not mean necessarily mean each value should be positive.


duckboy81:
Mistake in understanding how a calculated offset interacts with the bounding box.

PR #648, opened 2022-04-13
fix: add children to the DraggableCoreProps

### LINKED ISSUE #647: Incompatible with @types/react v18.x due to removal of children from FC and Component types
If you update [@types/react](https://www.npmjs.com/package/@types/react) to version 18 it introduces a breaking change that makes `<DraggableCore/>` fail typescript compilation.  This is due to the [removal of implicit children](https://solverfox.dev/writing/no-implicit-children/) from FC and Component.

>  Property 'children' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<DraggableCore> & Pick<Readonly<Partial<DraggableCoreProps>>, never> & Partial<...> & Partial<...>'.ts(2769)

SebastiaanNijland:
This should fix https://github.com/react-grid-layout/react-draggable/issues/647

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -43,6 +43,7 @@ declare module 'react-draggable' {
   export interface DraggableCoreProps {
     allowAnyClick: boolean,
     cancel: string,
+    children?: React.ReactNode,
     disabled: boolean,
     enableUserSelectHack: boolean,
     offsetParent: HTMLElement,


hasparus:
Until this PR is merged, the following code can serve as a workaround

```ts
import Draggable from "react-draggable";

declare module "react-draggable" {
  export interface DraggableProps {
    children: React.ReactNode;
  }
}
```

avdaredevil:
@STRML could we get this merged soon?

PR #680, opened 2022-11-02
feat: Support for using React Ref to specify handle and cancel

### LINKED ISSUE #679: Wanted: Supports passing in a Ref or HTMLElement as handle
It is easy to increase the mental burden by specifying the handle through the selector. I need to always pay attention that there is no duplicate name in these selectors. In React, Ref is very flexible and convenient. If it can support specifying handle through Ref, it must be very cool.

zhmushan:
close #679 

When I was writing this PR, I noticed that you used a very genius way to avoid possible problems caused by duplication of className.

However, the atomic class represented by Tailwind is becoming more and more popular, and many people no longer want to use className to do functional things, so I continue to complete this PR.

Also, this is my first time using this library and my first time using Flow, so please correct me if I'm doing something wrong.

Thanks again for your work, I have tried a lot of drag and drop libraries, `react-draggable` is the most suitable for my current product needs, thank you for saving me a lot of time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,10 +182,10 @@ axis: string,
 //   can be moved.
 bounds: {left?: number, top?: number, right?: number, bottom?: number} | string,
 
-// Specifies a selector to be used to prevent drag initialization. The string is passed to
+// Specifies a ref or selector to be used to prevent drag initialization. The selector is passed to
 // Element.matches, so it's possible to use multiple selectors like `.first, .second`.
 // Example: '.body'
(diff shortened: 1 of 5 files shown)


STRML:
Thank you for this contribution\! While the approach is technically sound, I'm hesitant to overload the string props (handle/cancel) with ref support. It adds complexity and bundle size without significant benefit for most users. If there's popular demand for this feature in the future, we can revisit. Closing for now.

PR #690, opened 2022-12-09
remove unnecessary global typescript namespace declaration

pgoldberg:
Declaring a global module with `declare module 'react-draggable'` is unnecessary, and can cause type pollution if multiple versions of `react-draggable` end up in a package.

This is exemplified in this repro: https://github.com/pgoldberg/react-draggable-typings-bug

The global module declaration is not necessary. Typescript will appropriately associate the typings with the `react-draggable` module, since the typings are declared in the package.json here: https://github.com/react-grid-layout/react-draggable/blob/44a8c6ed103ec6c0a4dda5faf7f8ebca16f9b325/package.json#L24

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1,67 +1,65 @@
-declare module 'react-draggable' {
-  import * as React from 'react';
+import * as React from 'react';
 
-  export interface DraggableBounds {
-    left?: number
-    right?: number
-    top?: number
-    bottom?: number
-  }
+export interface DraggableBounds {
+  left?: number
+  right?: number
+  top?: number
+  bottom?: number
+}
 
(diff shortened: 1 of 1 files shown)


STRML:
Closing this PR as stale. This TypeScript namespace cleanup PR has been open since 2022 without being merged. If this typing issue persists, please open a fresh PR against the current codebase. Thank you for the contribution!

STRML:
Reopening - this fixes a legitimate TypeScript type pollution issue.

gluxon:
Thanks @STRML!

I believe we ended up working around this issue with a patch on our side, so the issue does still exist. A fix for this would still be valuable!

STRML:
Closing in favor of #787 which implements this fix against the current codebase. Thank you for the contribution!

PR #699, opened 2023-01-16
Fix react18 concurrent mode drag glitch due to state inconsistency

### LINKED ISSUE #698: Fix React 18 Concurrent Mode drag glitches
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `react-draggable@4.4.5` for the project I'm working on.

In React 18 concurrent mode setState updates gets batched by default. That means they never run synchronously — always deferred.

I've faced an issue when drag parameters got out of sync after React 18 update on slow machine.
Visual manifestation were:

https://user-images.githubusercontent.com/2434839/212761083-648aca50-0b14-403b-82ad-2a778ebdc3d4.mov

items were dragged more far away than the cursor moved when CPU was throttled / occupied.

My investigation brought me to setState that is used for in-between drag events state and that `this.state` can and returns stale (old) values until rerender occurs in React 18.

As this state is not used inside render() function and updates do not need to trigger rerender, being motivated to make minimal change that would work for me and likely not break anything,

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/react-draggable/build/cjs/DraggableCore.js b/node_modules/react-draggable/build/cjs/DraggableCore.js
index 07b05f1..d9ea373 100644
--- a/node_modules/react-draggable/build/cjs/DraggableCore.js
+++ b/node_modules/react-draggable/build/cjs/DraggableCore.js
@@ -175,12 +175,7 @@ var DraggableCore = /*#__PURE__*/function (_React$Component) {
       // touchpoint was set to this element.
 
       var touchIdentifier = (0, _domFns.getTouchIdentifier)(e);
-
-      _this.setState({
-        touchIdentifier: touchIdentifier
-      }); // Get the current drag point from the event. This is used as the offset.
-
-
+      this.state.touchIdentifier = touchIdentifier; // Get the current drag point from the event. This is used as the offset.
       var position = (0, _positionFns.getControlPosition)(e, touchIdentifier, _assertThisInitialized(_this));
       if (position == null) return; // not possible but satisfies flow
 
@@ -201,11 +196,9 @@ var DraggableCore = /*#__PURE__*/function (_React$Component) {
       // so we know how much we've moved during the drag. This allows us
       // to drag elements around even if they have been moved, without issue.
 
-      _this.setState({
-        dragging: true,
-        lastX: x,
-        lastY: y
-      }); // Add events to the document directly so we catch when the user's mouse/touch moves outside of
+      _this.state.dragging = true;
+      _this.state.lastX = x;
+      _this.state.lastY = y; // Add events to the document directly so we catch when the user's mouse/touch moves outside of
       // this element. We use different events depending on whether or not we have detected that this
       // is a touch-capable device.
 
@@ -262,10 +255,8 @@ var DraggableCore = /*#__PURE__*/function (_React$Component) {
         return;
       }
 
-      _this.setState({
-        lastX: x,
-        lastY: y
-      });
+      _this.state.lastX = x
+      _this.state.lastY = y
     });
 
     _defineProperty(_assertThisInitialized(_this), "handleDragStop", function (e) {
@@ -303,11 +294,9 @@ var DraggableCore = /*#__PURE__*/function (_React$Component) {

erictheswift:
fixes https://github.com/react-grid-layout/react-draggable/issues/698

### DIFF
--- a/lib/DraggableCore.js
+++ b/lib/DraggableCore.js
@@ -298,7 +298,7 @@ export default class DraggableCore extends React.Component<DraggableCoreProps, D
     // distinguish between individual touches on multitouch screens by identifying which
     // touchpoint was set to this element.
     const touchIdentifier = getTouchIdentifier(e);
-    this.setState({touchIdentifier});
+    this.state.touchIdentifier = touchIdentifier;
 
     // Get the current drag point from the event. This is used as the offset.
     const position = getControlPosition(e, touchIdentifier, this);
@@ -322,12 +322,9 @@ export default class DraggableCore extends React.Component<DraggableCoreProps, D
     // Initiate dragging. Set the current x and y as offsets
     // so we know how much we've moved during the drag. This allows us
     // to drag elements around even if they have been moved, without issue.
-    this.setState({
-      dragging: true,
-
-      lastX: x,
-      lastY: y
-    });
+    this.state.dragging = true;
+    this.state.lastX = x;
+    this.state.lastY = y;
 
     // Add events to the document directly so we catch when the user's mouse/touch moves outside of
     // this element. We use different events depending on whether or not we have detected that this
@@ -372,10 +369,8 @@ export default class DraggableCore extends React.Component<DraggableCoreProps, D
       return;
     }
 
-    this.setState({
-      lastX: x,
-      lastY: y
-    });
+    this.state.lastX = x;
+    this.state.lastY = y;
   };
 
   handleDragStop: EventHandler<MouseTouchEvent> = (e) => {
--- a/lib/DraggableCore.js
+++ b/lib/DraggableCore.js
@@ -28,13 +28,6 @@ const eventsFor = {
 // Default to mouse events.
 let dragEventFor = eventsFor.mouse;
 
-type DraggableCoreState = {
-  dragging: boolean,
-  lastX: number,
-  lastY: number,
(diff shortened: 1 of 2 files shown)


maxwellwalin:
Confirmed that it solved the same issue I was having when I upgraded to React 18.

DeepAnchor:
This fixed an issue I was having when dragging an item very quickly across the screen. 

As in #698, the dragged item would move progressively further and further away from the cursor proportionate to the dragging acceleration (which makes sense since this issue is caused by React 18 batching).

Thanks @erictheswift for investigating and fixing this! 

LGTM

tnrich:
@STRML can we please get this on in? Thank you!

STRML:
Assigning directly to `state` is a pretty bad habit.

That said, you are correct that this data is not used for rendering, so it shouldn't be on `this.state` at all! Instead, it should simply held as local variables on the component.

If you make that change, I'll merge it.

erictheswift:
Agree)
My initial intent was to make patch be as minimum surface area as possible

erictheswift:
@STRML done

STRML:
Thanks!

trolund:
Hey! Just moved our code base to React 18 and i still have this problem even with `4.5.0` - Any ideas of why this is and how to fix it?

PR #751, opened 2024-05-08
Dragging elements in a rotated coordinate system

yanusa3:


thirstycode:
@yanusa3 are you using this forked rotated coordinate system? Is it production ready?

yanusa3:
> @yanusa3 are you using this forked rotated coordinate system? Is it production ready?

Yes, it's working for us already. We use it in cartography.

thirstycode:
Awesome. Thank you for this contribution!

README.md
```diff
@@ -278,6 +279,11 @@ positionOffset: {x: number | string, y: number | string},
 // you to, for example, get the correct drag deltas while you are zoomed in or out via
 // a transform or matrix in the parent of this element.
 scale: number
+
+// Specifies the angle of the canvas your are dragging this element on. This allows
+// you to, for example, get the correct drag deltas while you are rotated via
+// a transform or matrix in the parent of this element.
+angle: number
```
galangel, line 286: good name, to avoid confusion I would specify in the doc its in **degrees** since the CSS angle dataType can be set in degrees, gradians, radians, or turns

galangel:
why not add a description to the PR? or a gif?

yanusa3:
> why not add a description to the PR? or a gif?

Sorry, I forgot about this

README.md
yanusa3: This angle in degrees

lib/utils/positionFns.js
```diff
@@ -103,12 +103,13 @@ export function createCoreData(draggable: DraggableCore, x: number, y: number):
 // Create an data exposed by <Draggable>'s events
 export function createDraggableData(draggable: Draggable, coreData: DraggableData): DraggableData {
   const scale = draggable.props.scale;
+  const angleRadian = Math.PI * draggable.props.angle / 180.0;
   return {
     node: coreData.node,
-    x: draggable.state.x + (coreData.deltaX / scale),
```
STRML: This is nice; however the vast majority of usage of this library will have `angle: 0`, and I don't want to add overhead to every user just for this edge case. Can you clean this up so the code only runs if `draggable.props.angle` is 0?

STRML (changes requested):

lib/utils/positionFns.js
yanusa3: Certainly!

PR #754, opened 2024-05-15
fix/esm-rollup-clsx

carlblock:
Upgraded the clsx dependency and changed the default import of [clxs](https://www.npmjs.com/package/clsx) package to a named import.

This allows better compatibility with esm projects that uses rollup / vite.

Related to:

[fix: fix esm support and types](https://github.com/lukeed/clsx/pull/57)

[(0 , _clsx2.default) is not a function](https://github.com/react-grid-layout/react-draggable/issues/667)


galangel (approved):
lgtm

Issue #738, opened 2023-11-22
Draggable does not work as intended in 4.4.6. (Normal operation in 4.4.5)


Lee-Minhoon:
In version 4.4.6, the mouse leaves the draggable component.

**When not to use state**
```
import { useRef } from "react";
import Draggable from "react-draggable";
import "./App.css";

function App() {
  const nodeRef = useRef<HTMLDivElement>(null);

  return (
    <Draggable handle={`.handler`} nodeRef={nodeRef}>
      <div ref={nodeRef} style={{ width: "500px" }}>
        <div className="handler" style={{ backgroundColor: "gray" }}>
          React Draggable 4.4.6
        </div>
        <div>Body</div>
      </div>
    </Draggable>
  );
}

export default App;
```

**4.4.6**
https://github.com/react-grid-layout/react-draggable/assets/59780565/3594f5b5-9b4e-4659-aefd-6a5f60205037

**4.4.5**
https://github.com/react-grid-layout/react-draggable/assets/59780565/b7e8384d-1707-4f68-ad00-cc71affd61f0

**When to use states**
```
import { useCallback, useRef, useState } from "react";
import Draggable, { DraggableData, DraggableEvent } from "react-draggable";
import "./App.css";

function App() {
  const nodeRef = useRef<HTMLDivElement>(null);
  const [position, setPosition] = useState({ x: 0, y: 0 });

  const handleDrag = useCallback((e: DraggableEvent, data: DraggableData) => {
    setPosition({ x: data.x, y: data.y });
    console.log(data);
  }, []);

  return (
    <Draggable
      handle={`.handler`}
      nodeRef={nodeRef}
      position={position}
      onDrag={handleDrag}
    >
      <div ref={nodeRef} style={{ width: "500px" }}>
        <div className="handler" style={{ backgroundColor: "gray" }}>
          React Draggable 4.4.6
        </div>
        <div>Body</div>
      </div>
    </Draggable>
  );
}

export default App;
```

**4.4.6 with console log**
https://github.com/react-grid-layout/react-draggable/assets/59780565/a41c47d3-7793-4c68-bae1-f043297eaa22

**4.4.5 with console log**
https://github.com/react-grid-layout/react-draggable/assets/59780565/04ce9bb2-1aab-43a8-84ff-aea73a704178

AndreiAyar:
Seems related to this: [#2003](https://github.com/react-grid-layout/react-grid-layout/issues/2003)

[...]

PR #756, opened 2024-06-06
fix: use flushSync to update state

kaifeiji:
# Description

Just like https://github.com/react-grid-layout/react-grid-layout/pull/2043 says, `react-draggable` has the same problem from React 18's automatic batching -- cursor desync with dragging panel.

Though it's not quite recommended, but using flushSync is a quick and valid way to solve this. Further improvement needs some refactoring to be done.

# Related issues

https://github.com/react-grid-layout/react-draggable/issues/738

Please help have a look too. Thank you for your contribution. @STRML

### DIFF
--- a/lib/Draggable.js
+++ b/lib/Draggable.js
@@ -1,5 +1,6 @@
 // @flow
 import * as React from 'react';
+import { flushSync } from "react-dom";
 import PropTypes from 'prop-types';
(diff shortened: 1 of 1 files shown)


STRML:
#698 / #699 should have fixed this issue, not introduced it as #738 suggests.

kaifeiji:
Gotcha!

PR #760, opened 2024-08-08
fix: add allowMobileScroll prop to allow for clicks to optionally pass through on mobile

### LINKED ISSUE #728: Not possible to listen to click events inside a `Draggable` on Mobile
Hi,

we are using react-draggable and have a link inside a draggable which should be clickable.

```js
 <Draggable
        axis="x"
        >
        <Link>I should be clickable</Link>
  </Draggable>
```

On a non-touch device this works fine as the click event fires normally. However, on touch devices the click event is not fired at all, which is unfortunate.
The line of code responsible for this, seems to be this one:

https://github.com/react-grid-layout/react-draggable/blob/44a8c6ed103ec6c0a4dda5faf7f8ebca16f9b325/lib/DraggableCore.js#L295

`touchstart` event are cancelled there, resulting in the click never firing.

Since this line seems to be in place to prevent scrolling on touch devices, would a possible solution be to cancel touchmove instead?

Another solution may be to cancel the touchstart event higher up

tnrich:
Fixes https://github.com/react-grid-layout/react-draggable/issues/728

@STRML let me know if this PR looks good to you. Thanks!

Add doNotPreventMobileScroll prop to allow for clicks to optionally pass through on mobile

### DIFF
--- a/lib/DraggableCore.js
+++ b/lib/DraggableCore.js
@@ -288,7 +288,7 @@ export default class DraggableCore extends React.Component<DraggableCoreProps> {
 
     // Prevent scrolling on mobile devices, like ipad/iphone.
     // Important that this is after handle/cancel.
-    if (e.type === 'touchstart' && !this.props.doNotPreventMobileScroll) e.preventDefault();
+    if (e.type === 'touchstart' && !this.props.allowMobileScroll) e.preventDefault();
 
     // Set touch identifier in component state if this is a touch event. This allows us to
     // distinguish between individual touches on multitouch screens by identifying which
--- a/lib/DraggableCore.js
+++ b/lib/DraggableCore.js
@@ -288,7 +288,7 @@ export default class DraggableCore extends React.Component<DraggableCoreProps> {
 
     // Prevent scrolling on mobile devices, like ipad/iphone.
     // Important that this is after handle/cancel.
-    if (e.type === 'touchstart') e.preventDefault();
+    if (e.type === 'touchstart' && !this.props.doNotPreventMobileScroll) e.preventDefault();
 
     // Set touch identifier in component state if this is a touch event. This allows us to
     // distinguish between individual touches on multitouch screens by identifying which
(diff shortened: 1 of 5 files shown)


STRML:
Can you please invert it: "preventMobileScroll", default true.

tnrich:
What about "allowMobileScroll"? I prefer having the default be false @STRML

lib/DraggableCore.js
```diff
@@ -288,7 +288,7 @@ export default class DraggableCore extends React.Component<DraggableCoreProps> {
 
     // Prevent scrolling on mobile devices, like ipad/iphone.
     // Important that this is after handle/cancel.
-    if (e.type === 'touchstart') e.preventDefault();
+    if (e.type === 'touchstart' && !this.props.doNotPreventMobileScroll) e.preventDefault();
```
STRML: Still needs change

STRML (changes requested):

lib/DraggableCore.js
tnrich: whoops not sure how that didn't get updated. thanks!

STRML (approved):

STRML:
Thanks for the contribution. Please see the additional commit for fixups.

tnrich:
Thanks @STRML !! Much appreciated 🙏

specs/draggable.spec.jsx
```diff
@@ -622,6 +622,18 @@ describe('react-draggable', function () {
       assert.equal(drag.state.dragging, true);
     });
 
+    it('should *not* call preventDefault on touchStart event if "allowMobileScroll"', function () {
+      drag = TestUtils.renderIntoDocument(<Draggable allowMobileScroll={true}><div/></Draggable>);
+
+      const e = new Event('touchstart');
+      // Oddly `e.defaultPrevented` is not changing here. Maybe because we're not mounted to a real doc?
+      let pdCalled = false;
```
tnrich, line 630: nice one @STRML , thanks for adding a test! my Achilles heel!

tnrich:
@STRML it appears that I need a new release in order to be able to use the updated code. Do you think you could do that plz! 

Thanks! 

image

taniaprocha:
When do we have the next release to use this `allowMobileScroll` prop?

consistent-k:
@STRML @tnrich Which version supports the use of this new prop?

imrahil:
Any chances to release the new version with this fix?

PR #763, opened 2024-09-06
Fix bounds="selector" functionality when in a Shadow DOM tree.

miguelcalderon:
Hi! I'm Miguel Calderón and work at PSPDFKit. I'd like to say thanks for your package, which is great!

We've recently switched our SDK container from Iframe to Shadow DOM, and are about to upgrade to React 18, which is bringing up lots of small headaches, mostly caused by our own tech debt.

One of the problems we're finding is using `react-draggable` from within the Shadow DOM when setting the `bounds` prop to a selector.

## Details

The `bounds` prop was not working correctly when set to a selector within a Shadow DOM tree. The reason is that the provided selector was being queried to `ownerDocument` and would throw if the node was attached within a Shadow root.

Here's an example showing the problem: https://playcode.io/1996274

- In the above playground try to drag the square - it will log an error instead.
- If you set `withShadow` to `false` it will work.

The solution applied in this PR is to query the node returned by `node.getRootNode()` instead: this method returns `ownerDocument` when called on the normal DOM as before, but when the node is inside a Shadow root, the shadow root will be returned instead. Additionally, if the node is inside a fragment and not attached to a document nor a Shadow root, the topmost element will be returned.

Some tests have been added as well to cover `bounds` as `parent` as selector, from within a Shadow root and in the normal DOM.

### DIFF
--- a/lib/utils/positionFns.js
+++ b/lib/utils/positionFns.js
@@ -22,8 +22,14 @@ export function getBoundPosition(draggable: Draggable, x: number, y: number): [n
     if (bounds === 'parent') {
       boundNode = node.parentNode;
     } else {
-      boundNode = ownerDocument.querySelector(bounds);
+      // Flow assigns the wrong return type (Node) for getRootNode(),
+      // so we cast it to one of the correct types (Element).
+      // The others are Document and ShadowRoot.
+      // All three implement querySelector() so it's safe to call.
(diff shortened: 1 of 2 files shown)


galangel:
lgtm

STRML:
The kind of PR I like to see - very few changes, very many tests! Thank you.

PR #773, opened 2025-01-15
Fix forgotten requestAnimationFrame call

JefHellemansBlockbax:
In PR https://github.com/react-grid-layout/react-draggable/pull/556, a performance issue with a forced reflow was fixed, but there was a second case that was forgotten. I've applied the same fix to that case and extracted a scheduling helper so this doesn't accidentally happen again.


galangel (approved):

STRML:
Thanks, very helpful.

PR #791, opened 2026-01-06
Fix CSP warning

yashjais:
**Summary**
This PR adds Content Security Policy (CSP) support to react-draggable by introducing a nonce prop that can be applied to dynamically created style elements. 

Fixes this [issue](https://github.com/react-grid-layout/react-draggable/issues/659)

**Changes**

**Core Implementation**
- Added nonce prop to DraggableCore component that accepts an optional string value
- Updated addUserSelectStyles function in lib/utils/domFns.js to accept and apply the nonce attribute to dynamically created <style> tags
- Updated TypeScript definitions to include the new nonce prop in DraggableCoreProps

**Documentation**
- Updated README.md to document the new nonce prop in both:
- The main props table with description: "Nonce for Content Security Policy. Applied to dynamically created <style> tags"
- The inherited props list for the Draggable component

**Testing**
- Added comprehensive test coverage for the nonce functionality:
- Test verifying that nonce attribute is correctly applied when prop is provided
- Test confirming that nonce attribute is not added when prop is omitted
- Both tests include proper cleanup of existing style elements

**Use Case**
- This feature enables react-draggable to work in environments with strict CSP policies that require nonces for inline styles. When enableUserSelectHack is enabled, the library dynamically creates a <style> tag to prevent text selection during dragging. With this change, developers can now provide a CSP nonce to make this compliant with their security policies.

**Example Usage**
```
<DraggableCore nonce="random-nonce-value">
  <div>Drag me</div>
</DraggableCore>
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,7 @@ type DraggableData = {
 | `grid` | `[number, number]` | - | Snap to grid `[x, y]` |
 | `handle` | `string` | - | CSS selector for the drag handle |
 | `nodeRef` | `React.RefObject` | - | Ref to the DOM element. Required for React Strict Mode |
+| `nonce` | `string` | - | Nonce for Content Security Policy. Applied to dynamically created `<style>` tags |
 | `offsetParent` | `HTMLElement` | - | Custom offsetParent for drag calculations |
 | `onDrag` | `DraggableEventHandler` | - | Called while dragging |
 | `onMouseDown` | `(e: MouseEvent) => void` | - | Called on mouse down |
@@ -147,6 +148,7 @@ See [React-Resizable](https://github.com/react-grid-layout/react-resizable) and 
 - `handle`
 - `nodeRef`
 - `offsetParent`
+- `nonce`
 - `onDrag`
 - `onMouseDown`
 - `onStart`
--- a/lib/DraggableCore.js
+++ b/lib/DraggableCore.js
@@ -50,6 +50,7 @@ export type DraggableCoreDefaultProps = {
   onStop: DraggableEventHandler,
   onMouseDown: (e: MouseEvent) => void,
   scale: number,
+  nonce: ?string,
 };
 
 export type DraggableCoreProps = {
@@ -60,6 +61,7 @@ export type DraggableCoreProps = {
   grid: [number, number],
   handle: string,
   nodeRef?: ?React.ElementRef<any>,
+  nonce?: ?string,
 };
 
 //
@@ -214,6 +216,12 @@ export default class DraggableCore extends React.Component<DraggableCoreProps> {
     scale: PropTypes.number,
 
     /**
+     * `nonce`, if set, will be added to the dynamically created <style> tag
+     * for Content Security Policy compliance
+     */
+    nonce: PropTypes.string,
+
+    /**
      * These properties should be defined on the child, not here.
      */
     className: dontSetMe,
@@ -231,6 +239,7 @@ export default class DraggableCore extends React.Component<DraggableCoreProps> {
     onStop: function(){},
     onMouseDown: function(){},
     scale: 1,
+    nonce: undefined,
   };
 
   dragging: boolean = false;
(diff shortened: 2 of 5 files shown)


yashjais:
@STRML Could you please take a look?

STRML:
Thanks for working on this! However, I think there's a simpler solution that doesn't require any library changes.

## Alternative: Pre-create the style element server-side

Looking at the code, `addUserSelectStyles` checks for an existing element before creating one:

```javascript
let styleEl = doc.getElementById('react-draggable-style-el');
if (!styleEl) {
  // ... creates the element
}
```

**Users can simply add this style tag server-side with their CSP nonce:**

```html
<style id="react-draggable-style-el" nonce="YOUR_NONCE_HERE">
.react-draggable-transparent-selection *::-moz-selection {all: inherit;}
.react-draggable-transparent-selection *::selection {all: inherit;}
</style>
```

When react-draggable runs, it will find the existing element and skip creating its own - no CSP violation.

## Benefits of this approach

1. **No library changes needed** - works today with current versions
2. **No prop threading** - don't need to pass nonce through component tree
3. **Server controls the nonce** - as CSP intended
4. **Static, cacheable** - the CSS content never changes

This could be documented in the README as the recommended CSP solution rather than adding API surface to the library.

What do you think?

yashjais:
Thank you for the alternative solution. We're using static pages, so we implemented your approach by running this function at the top level of our app:

> 
> /**
>  * Utility to inject react-draggable style element with CSP nonce support.
>  * 
>  * This function implements STRML's suggested solution for CSP compliance:
>  * Pre-creates the style element that react-draggable would create, but with
>  * the CSP nonce attribute. When react-draggable runs, it will find this
>  * existing element and skip creating its own, avoiding CSP violations.
>  */
> export const injectReactDraggableStyle = () => {
>   // Check if the style element already exists
>   const existingStyle = document.getElementById('react-draggable-style-el');
>   if (existingStyle) {
>     // Element already exists, no need to create it
>     return;
>   }
> 
>   // Get the nonce from the meta tag (works for both static and dynamic nonces)
>   const nonceMeta = document.querySelector('meta[name="csp-nonce"]');
>   const nonce = nonceMeta?.getAttribute('content') || '';
> 
>   // Create the style element with the exact ID that react-draggable looks for
>   const styleEl = document.createElement('style');
>   styleEl.id = 'react-draggable-style-el';
>   
>   // Apply nonce if available
>   if (nonce) {
>     styleEl.setAttribute('nonce', nonce);
>   }
> 
>   // Set the CSS content that react-draggable would inject
>   styleEl.textContent = \`
>     .react-draggable-transparent-selection *::-moz-selection {
>       all: inherit;
>     }
>     .react-draggable-transparent-selection *::selection {
>       all: inherit;
>     }
>   \`;
> 
>   // Insert into the document head
>   document.head.appendChild(styleEl);
> };

We call this function in our index.js before React renders:

> import { injectReactDraggableStyle } from './utils/cspNonceHelper';
> 
> // Inject react-draggable style element with CSP nonce before React renders
> injectReactDraggableStyle();
> 
> const root = ReactDOM.createRoot(document.getElementById('root'));
> // ... rest of app initialization

This works well for our static page setup. The function reads the nonce from a meta tag in our HTML, so it supports both static and dynamic nonces depending on how the meta tag is populated.

Thanks again for the suggestion!

nhristova:
Our app does not use server side rendering. 
We were able to compensate for the missing style by adding it to our css file. 
We still get the CSP error in the browser console.

```css
.react-draggable-transparent-selection *::selection {
    all: inherit;
}
.react-draggable-transparent-selection *::-moz-selection {
    all: inherit;
}
```

I do not understand why the users of the library need to implement workarounds for CSP blocking style tags added dynamically to the document.

Passing nonce as a prop seems to me would have been the standard approach.

STRML:
@nhristova you're right. Fixed this.

PR #810, opened 2026-06-29
fix: process is not defined in browser environment (#806)

### LINKED ISSUE #806: process.env.DRAGGABLE_DEBUG referenced in browser bundle (v4.6.0)
**Version:** 4.6.0  
**Environment:** Browser (Vite / any non-webpack bundler)

### Problem

`build/cjs/chunk-D5BXCJ5G.mjs` (and `build/cjs/Draggable.js`) ship a raw `process.env.DRAGGABLE_DEBUG` reference that is never replaced at build time:

```js
// log.ts → chunk-D5BXCJ5G.mjs:343
if (process.env.DRAGGABLE_DEBUG) console.log(...args);
```

`process` is a Node.js global. Bundlers like Vite do **not** inject it into browser bundles, so any app that imports react-draggable crashes on load with:

```
ReferenceError: process is not defined
    at log (chunk-D5BXCJ5G.mjs:343)
    at getDerivedStateFromProps (chunk-D5BXCJ5G.mjs:759)
    ...
```

This is a regression — prior releases did not reference `process.env` in the browser bundle.

### Expected behaviour

The published browser bundle should not reference `process.env`. Either:
- replace the reference at build time (e.g. `if (false)` in the dist), or
- use `typeof process !== 'undefined' && process.env.DRAGGABLE_DEBUG` as a guard.

### Workaround

Add a `define` entry in `vite.config.ts` to stub the variable:

```ts
define: {
  "process.env.DRAGGABLE_DEBUG": JSON.stringify(process.env.DRAGGABLE_DEBUG ?? false),
},
```

This is a consumer-side workaround and should not be required.

xfournet:
Fix #806 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved debug logging behavior so it safely handles environments where `process` is unavailable.
  * Logging now consistently falls back to a no-op when debug mode is not enabled, reducing unnecessary output.

### DIFF
--- a/lib/utils/log.ts
+++ b/lib/utils/log.ts
@@ -1,4 +1,3 @@
 /*eslint no-console:0*/
-export default function log(...args: unknown[]): void {
-  if (process.env.DRAGGABLE_DEBUG) console.log(...args);
-}
+const log = typeof process !== 'undefined' && process.env.DRAGGABLE_DEBUG ? console.log.bind(console) : function noop(): void {};
+export default log;


johnhunter:
Any plans to release this?

PR #812, opened 2026-07-27
fix: prevent process.env access from crashing browser builds

Kanhaiyaarora:
## Summary

This PR prevents runtime errors in browser environments by safely checking for the existence of `process` before accessing `process.env.DRAGGABLE_DEBUG`.

## Problem

The current implementation directly accesses `process.env.DRAGGABLE_DEBUG`:

```ts
if (process.env.DRAGGABLE_DEBUG) console.log(...args);
```

In browser environments where `process` is not defined (e.g., Vite without a process polyfill), this can lead to runtime errors and affect components that depend on `react-draggable`.

## Solution

Added a guard to ensure `process` exists before accessing `process.env`:

```ts
if (typeof process !== "undefined" && process.env?.DRAGGABLE_DEBUG) {
  console.log(...args);
}
```

This preserves the existing debug logging behavior while preventing errors in environments where `process` is unavailable.

## Testing

* Verified the issue in a React + Vite application.
* Confirmed that the runtime error no longer occurs.
* Confirmed that drag functionality works correctly after the change.
* Verified that debug logging behavior remains unchanged when `DRAGGABLE_DEBUG` is enabled.
image

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility in environments without a `process` object.
  * Debug logging now appears only when explicitly enabled through the debug setting.

### DIFF
--- a/lib/utils/log.ts
+++ b/lib/utils/log.ts
@@ -1,4 +1,6 @@
 /*eslint no-console:0*/
 export default function log(...args: unknown[]): void {
-  if (process.env.DRAGGABLE_DEBUG) console.log(...args);
+  if (typeof process !== "undefined" && process.env?.DRAGGABLE_DEBUG) {
+    console.log(...args);
+  }
 }


STRML:
Thanks for this. Closing as a duplicate of #810, which landed in 24a8fb3 and fixes the same crash.

For the record on why I took that one: the `?.` in `process.env?.DRAGGABLE_DEBUG` defeats webpack's `EnvironmentPlugin` substitution in the UMD build. I built both variants to check. With optional chaining, `build/web/react-draggable.min.js` keeps a live `process` reference:

```js
process !== "undefined" && ((_a = process.env) == null ? void 0 : _a.DRAGGABLE_DEBUG)
```

With the plain member access it substitutes and dead-code eliminates:

```js
process !== "undefined" && false
```

The `typeof process` guard is the part that actually fixes #806, and you and @xfournet arrived at it separately. Appreciate you filing it.

PR #817, opened 2026-09-03
fix: type each `exports` condition so ESM consumers get ESM types

### LINKED ISSUE #816: Types: `exports` map points both conditions at the CJS declaration — default import unusable under `moduleResolution: NodeNext` (4.6.0+)
### Summary

Since the tsup build landed in 4.6.0, the `exports` map declares a single top-level `"types"` entry that sits **outside** the `import`/`require` conditions:

```jsonc
// package.json (4.7.1)
"exports": {
  ".": {
    "types": "./build/cjs/cjs.d.ts",
    "import": "./build/cjs/cjs.mjs",
    "require": "./build/cjs/cjs.js"
  },
  "./package.json": "./package.json"
}
```

Both conditions therefore resolve to `build/cjs/cjs.d.ts`. The package has no `"type": "module"`, so that file is a **CommonJS** declaration — and under `moduleResolution: "node16" | "nodenext"` a default import of a CJS module is typed as the whole `module.exports` object rather than the declared default export. The result is that `Draggable` cannot be used as a JSX component.

`build/cjs/cjs.d.mts` — the correct ESM declaration — is shipped in the tarball, but nothing in the `exports` map can reach it.

This is types-only; the runtime is fine (`import D from "react-draggable"` gives `function Draggable` under Node ESM).

### Versions

- react-draggable 4.7.1 (introduced in 4.6.0, which added the `exports` map)
- typescript 6.0.3 (also reproduces on 5.x)
- @types/react 19.2.18

### Reproduction

`package.json`

```json
{ "name": "repro", "type": "module", "version": "1.0.0" }
```

`tsconfig.json`

```json
{
  "compilerOptions": {
    "module": "NodeNext",
    "moduleResolution": "NodeNext",
    "target": "ES2022",
    "jsx": "react-jsx",
    "esModuleInterop": true,
    "strict": true,
    "noEmit": true
  },
  "include": ["index.tsx"]
}
```

`index.tsx`

```tsx
import Draggable from "react-draggable";

export const App = () => (
	<Draggable>
		<div>drag me</div>
	</Draggable>
);
```

`npx tsc -p tsconfig.json`:

```
index.tsx(4,3): error TS2604: JSX element type 'Draggable' does not have any construct or call signatures.
index.tsx(4,3): error TS2786: 'Draggable' cannot be used as a JSX component.
  Its type 'typeof import(".../react-draggable/build/cjs/cjs")' is not a valid JSX element type.
```

`--traceResolution` confirms the ESM import is typed by the CJS declaration:

```
======== Module name 'react-draggable' was successfully resolved to
'.../react-draggable/build/cjs/cjs.d.ts' with Package ID
'react-draggable/build/cjs/cjs.d.ts@4.7.1'. ========
```

### Suggested fix

Move `types` inside each condition so the ESM entry gets the `.d.mts` that is already being built:

```diff
 "exports": {
   ".": {
-    "types": "./build/cjs/cjs.d.ts",
-    "import": "./build/cjs/cjs.mjs",
-    "require": "./build/cjs/cjs.js"
+    "import": {
+      "types": "./build/cjs/cjs.d.mts",
+      "default": "./build/cjs/cjs.mjs"
+    },
+    "require": {
+      "types": "./build/cjs/cjs.d.ts",
+      "default": "./build/cjs/cjs.js"
+    }
   },
   "./package.json": "./package.json"
 }
```

The root `main`/`module`/`types`/`typings` fields can stay as they are for legacy resolvers. I verified this change locally as a package patch — it clears both errors, with no runtime change.

`attw` agrees. Published 4.7.1:

```
🎭 Import resolved to a CommonJS type declaration file, but an ESM JavaScript file. (FalseCJS)

                    "react-draggable"
 node10             🟢
 node16 (from CJS)  🟢 (CJS)
 node16 (from ESM)  🎭 Masquerading as CJS
 bundler            🟢
```

With the `exports` change above:

```
                    "react-draggable"
 node10             🟢
 node16 (from CJS)  🟢 (CJS)
 node16 (from ESM)  🟢 (ESM)
 bundler            🟢
```

Bundler resolution is unaffected, which is presumably why this has not come up more often — it only bites `node16`/`nodenext` consumers.

### Possibly related: `DraggableEventHandler` parameter type

In 4.5.0 the handler took the library's own event union:

```ts
export type DraggableEventHandler = (e: DraggableEvent, data: DraggableData) => void | false;
```

In 4.7.1 it takes the **global DOM** `MouseEvent`:

```ts
type DraggableEventHandler = (e: MouseEvent, data: DraggableData) => void | false;
type DraggableEvent = React.MouseEvent<HTMLElement | SVGElement> | React.TouchEvent<HTMLElement | SVGElement> | MouseEvent | TouchEvent;
```

`onTouchStart` routes touch events into `handleDragStart`, which calls `props.onStart`, so a handler can be invoked with a `TouchEvent` — and with a React synthetic event when the drag starts from the React-attached `onMouseDown`. The declared parameter type excludes both, while `DraggableEvent` right above it describes them exactly. If this was an unintended consequence of the Flow → TypeScript migration, `DraggableEventHandler` taking `DraggableEvent` would match both the old surface and the runtime behaviour.

Happy to open a PR for either or both if the direction looks right.

binhpv:
Fixes #816.

### The bug

`exports` declared a single condition-independent `"types"`, so the `import` and the `require` entry were both typed by `build/cjs/cjs.d.ts`:

```jsonc
".": {
  "types": "./build/cjs/cjs.d.ts",
  "import": "./build/cjs/cjs.mjs",
  "require": "./build/cjs/cjs.js"
}
```

The package has no `"type": "module"`, so that declaration is CommonJS. Under `moduleResolution: node16`/`nodenext` a default import of a CJS module is typed as the whole `module.exports` object, so `<Draggable>` fails with TS2604/TS2786 and `attw` reports the ESM entry as masquerading as CJS. Regression in 4.6.0, which introduced the map. `build/cjs/cjs.d.mts` was already being emitted — nothing could reach it.

### The fix

Each condition declares its own `types`. Types-only: the `import`/`require` runtime targets are byte-for-byte what they were.

`attw` on a packed tarball, before → after:

```
 node16 (from ESM)  🎭 Masquerading as CJS      node16 (from ESM)  🟢 (ESM)
```

All four resolution modes are green after the change (`No problems found 🌟`); `node10`, `node16 (from CJS)` and `bundler` were green before and are unchanged.

### Why the existing checks missed it

`test/typeCompat` and `typings/` both map `react-draggable` to `lib/` source through a tsconfig `paths` entry under `moduleResolution: node`. `paths` bypasses `exports` entirely, so no existing check ever resolves the package the way a consumer does.

So `scripts/verify-build.cjs` gets a Contract 5, in two halves:

1. **Static** — `exports["."]` must not carry a condition-independent `types`; each condition must declare its own, pointing at the right extension, and every target must exist.
2. **End-to-end** — a temp ESM consumer imports `react-draggable` **by name** under `nodenext`, resolved through a `node_modules` symlink rather than a `paths` mapping, and must type-check against the built package.

Each half fails on the pre-fix map on its own — I checked the second by pointing `import.types` back at the `.d.ts` with the static block stripped, and it reproduces the original TS2604/TS2786.

### Verified

`make lint`, `make test` (204 passing), `make build` including the new contract, and `attw` as above. Browser tests were not run — Puppeteer's Chrome download failed in my environment.

### Note

I put the changelog entry under an `### Unreleased` heading since there's no unreleased-section convention here; fold it into the next version heading if you'd rather.

The other half of #816 — `DraggableEventHandler` taking the global `MouseEvent` while `DraggableEvent` right above it describes the React and touch events `onTouchStart` actually produces — is deliberately not in this PR. It changes the public type surface that `test/typeCompat/fixture.tsx` pins, so it seemed worth your call first. Happy to follow up with it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TypeScript type resolution for ESM consumers using Node16 or NodeNext module resolution.
  * Package imports now select the appropriate type declarations for ESM and CommonJS usage.

* **Tests**
  * Added build validation for ESM type resolution and package export configuration.

* **Documentation**
  * Corrected the heading level for the Unreleased changelog section.


binhpv:
@STRML Can you please take a look? Thank you in advance.

